PR #3, opened 2024-11-15
Andrew moj

andrewli2403:
- moj capability into the public facing library
- added customization features with protection checks for system prompt and conversation


SecroLoL:
LGTM!

PR #11, opened 2024-11-29
REFACTOR: Fix syntax for run_eval

SecroLoL:
Fixes syntax for running evals by uniformly using a judge model for all LLM judge scorers

Now, you can run all evals with 

```
results = client.run_evaluation(
        examples=[example1, example2],
        scorers=[scorer, c_scorer],
        model="QWEN",
        metadata={"batch": "test"},
        eval_run_name=eval_run_name,
        log_results=True,
    )
```

where we don't have to specify `c_scorer` to have a definition like 
```
c_scorer = CustomFaithfulnessMetric(
        threshold=0.6,
        model="gpt-4o-mini"
    )
```
and we just have
```
c_scorer = CustomFaithfulnessMetric(
        threshold=0.6,
    )
```

### DIFF
--- a/docs/prompt_scorer.ipynb
+++ b/docs/prompt_scorer.ipynb
@@ -15,18 +15,9 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "metadata": {},
-   "outputs": [
-    {
-     "name": "stderr",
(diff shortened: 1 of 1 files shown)


JCamyre:
Added clarifications and code changes, take a look when you get the chance

PR #15, opened 2024-12-08
Address empty ScorerData fields in ScoringResult

SecroLoL:
In the return object of `run_evalution()`, it may be possible that the `ScorerData` field in one or more of the `ScoringResult` objects is `[]`, which can be caused by the `Example` object not containing the correct fields specified by the scorers they run the evaluation with.

This PR adds a helper function that takes the processed `ScoringResult` objects and checks if they are missing `ScorerData` fields. If so, an error message is raised, but the overall evaluation is not canceled/no exception is raised.

This is because if we raised an error, then it may be annoying for people running evals with datasets where a single example lacks the fields.

### DIFF
--- a/judgeval/common/tracer.py
+++ b/judgeval/common/tracer.py
@@ -1,0 +1,159 @@
+"""
+Tracing system for judgeval that allows for function tracing using decorators.
+"""
+
+import time
+import functools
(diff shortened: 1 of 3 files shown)


andrewli2403:
lgtm

PR #32, opened 2025-01-06
Add Span Level Evals (multi-step evaluation)

JCamyre:
Adds Span level evals, enabling customers to evaluate any step in their multi-step workflow.


JCamyre:
> LGTM, what a sick PR

Thanks :)

PR #38, opened 2025-01-15
Add Custom Judge Models for Custom Scorers

SecroLoL:
Make it so that people can use custom judge models when using custom scorers.

NOTE: this is ONLY for custom scorers! Default scorers must be routed through ourselves


SecroLoL:
closing because old UTs are buggy, re-opening so the new CI tests execute

PR #58, opened 2025-02-07
Add unit testing docs (with Pytest)

SecroLoL:
Adds page to the dev docs with PyTest integration


JCamyre:
One UT didn't pass:
image

PR #62, opened 2025-02-11
Tracer Async Evaluations

JCamyre:
This PR makes evaluations asynchronous - meaning we are now enabling our customers to run their systems with no overhead while running evaluations for them!
This is super crucial, as who realistically wants to use an eval framework and platform that slows down their system?
The main change I made was saving EvaluationRun objects in the TraceData.entries . Previously, we were running the evaluations while the workflow was running (meaning we had to temporarily stop the workflow to run the evaluation). We then stored the results in EvalResult objects in TraceData.entries.
I accomplished this by creating a new server called "async eval worker" and using a queue called RabbitMQ.
When a trace is saved, we send the TraceData to the queue (essentially a FIFO queue that is deployed on AWS and can hold 100,000's of TraceData's). The async eval worker server then pulls TraceData's from the queue and calls the judgment backend server to run the evaluations from the EvaluationRun objects in the TraceData.entries.
The async eval worker receives the EvaluationResult's, then puts it in the corresponding TraceData.entries , then calls the judgment backend server again to save the updated TraceData (with the EvaluationResult's) to the traces table in the Supabase.
TLDR:
This PR makes evaluations asynchronous - meaning we are now enabling our customers to run their systems with no overhead while running evaluations for them!
This is super crucial, as who realistically wants to use an eval framework and platform that slows down their system?
This means you should no longer do: await judgment.get_current_trace().async_evaluate(...) .
Instead do: judgment.get_current_trace().async_evaluate(...) .


JCamyre (commented):
Run UT's

SecroLoL:
UTs failed, no `pika` module installed.

JCamyre (commented):
Rerun UT's

SecroLoL (approved):
LGTM. great work!

PR #74, opened 2025-02-17
Delete Trace and EvalRun endpoints

alanzhang25:


JCamyre:
image
I got this error while running `test_tracer.py`

JCamyre:
It would be nice to run this as a `pytest`, can you set up the trace_manager as a fixture, thanks.
image
I get this error when running via `pytest` because the fixture is not set up.

src/judgeval/judgment_client.py
```diff
@@ -194,6 +194,51 @@ def pull_eval(self, project_name: str, eval_run_name: str) -> List[Dict[str, Uni
             eval_run_result[0]["id"] = result_id
             eval_run_result[0]["results"] = [ScoringResult(**filtered_result)]
         return eval_run_result
+    
+    def delete_eval(self, project_name: str, eval_run_name: str) -> bool:
```
SecroLoL, line 198: Changed this function name; previous one was too verbose.

This is a client, so the func names need to be as short as possible and dev-friendly

src/judgeval/judgment_client.py
```diff
@@ -194,6 +194,51 @@ def pull_eval(self, project_name: str, eval_run_name: str) -> List[Dict[str, Uni
...
+        response = requests.delete(JUDGMENT_EVAL_DELETE_API_URL, 
+                        json=eval_run_request_body.model_dump(),
+                        headers={
+                            "Content-Type": "application/json",
+                        })
+        if response.status_code != requests.codes.ok:
+            raise ValueError(f"Error deleting eval results: {response.json()}")
+        return response.json()
+    
+    def delete_project_evals(self, project_name: str) -> bool:
```
SecroLoL, line 221: Same here

PR #75, opened 2025-02-19
Update docs with API reference

SecroLoL:
API References for certain frequently used of the external API
- `client.run_evaluation()`
- `Tracer` object


SecroLoL (commented):
LGTM

PR #79, opened 2025-02-21
LangGraph x Judgeval Integration

JCamyre:


src/judgeval/common/tracer.py
```diff
@@ -255,16 +271,25 @@ def record_evaluation(self, results: List[ScoringResult], start_time: float):
         if self._current_span:
             duration = time.time() - start_time  # Calculate duration from start_time
             
+            # print(f"self._current_span: {type(self._current_span)=}, {self._current_span=}")
```
SecroLoL, line 274: remove these print statements

src/judgeval/common/tracer.py
```diff
@@ -353,12 +378,15 @@ def condense_trace(self, entries: List[dict]) -> List[dict]:
                 current_entry["duration"] = entry["timestamp"] - current_entry["timestamp"]
                 condensed.append(current_entry)
                 active_functions.remove(function)
-                del function_entries[function]
+                # del function_entries[function]
                 
-            elif function in active_functions:
+            # The OR condition is to handle the Langchain LLM evaluation case.
```
SecroLoL, line 383: add more elaboration on what the LLM evaluation case is, I think this change deserves a long comment

src/judgeval/common/tracer.py
```diff
@@ -654,3 +709,193 @@ def _format_output_data(client: ApiClient, response: Any) -> dict:
             "total_tokens": response.usage.input_tokens + response.usage.output_tokens
         }
     }
+
+class JudgevalCallbackHandler(BaseCallbackHandler):
+    def __init__(self, trace_client: TraceClient):
+        self.trace_client = trace_client
+        self.openai_count = 1
+
+    def start_span(self, name: str, span_type: SpanType = "span"):
```
SecroLoL, line 718: I assume these are helpers, not part of the actual CallbackHandler, but rather some other component?

src/judgeval/common/tracer.py
```diff
@@ -654,3 +709,193 @@ def _format_output_data(client: ApiClient, response: Any) -> dict:
...
+            function=name,
+            depth=self.trace_client.tracer.depth,
+            message=f"← {name}",
+            timestamp=time.time(),
+            duration=duration,
+            span_type=span_type
+        ))
+        self.trace_client._current_span = self.trace_client.prev_span
+
+    def on_retriever_start(
```
SecroLoL, line 752: I think the `on_xyz` is the real triggered parts of the callback handler, right?

src/judgeval/common/tracer.py
```diff
@@ -654,3 +709,193 @@ def _format_output_data(client: ApiClient, response: Any) -> dict:
...
+        *,
+        run_id: UUID,
+        parent_run_id: Optional[UUID] = None,
+        tags: Optional[list[str]] = None,
+        metadata: Optional[dict[str, Any]] = None,
+        **kwargs: Any,
+    ) -> Any:
+        pass
+
+    def on_retriever_end(
```
SecroLoL, line 765: Why aren't we tracing retrievers? Isn't this important

src/judgeval/common/tracer.py
```diff
@@ -654,3 +709,193 @@ def _format_output_data(client: ApiClient, response: Any) -> dict:
...
+            'args': prompts,
+            'kwargs': kwargs
+        })
+
+    def on_llm_end(self, response: LLMResult, *, run_id: UUID, parent_run_id: Optional[UUID] = None, **kwargs: Any):
+        # print(f"LLM end: {response}")
+        self.trace_client.record_output(response.generations[0][0].text)
+        self.end_span(self.trace_client._current_span, span_type="tool")
+
+    def on_chat_model_start(
```
SecroLoL, line 874: I wonder if this needs to be generalized/add support for more providers; oh well, for now we can just assume that OpenAI, anthropic, and Together are the predominant inference providers

src/judgeval/common/tracer.py
SecroLoL, line 1: just needs some more cleaning (removing print statements, fleshing out the rest of the callback handler)

SecroLoL (changes requested):
Some comments mainly on the `tracer.py` file, seems like there's still some work to be done?

SecroLoL:
UTs failing due to an import issue with `langchain_core` package. Probably some kinda environment issue?

JCamyre:
Thanks for the comments, these are good. Not quite done yet - adding a more robust to properly check the span evaluation is for a LangChain LLM client

src/judgeval/common/tracer.py
JCamyre: `start_span` and `end_span` are helper funcs to work with the CallbackHandlers

src/judgeval/common/tracer.py
JCamyre: Yep

PR #82, opened 2025-02-25
New Example ID

Mandolaro:
Added new example_id associated with the evaluations. New example ID is UUID string.

Refactor old example_id into example_index (better suited with the use case)

Example of new example object:
[
        "examples": [
          {
            "name": null,
            "input": "What if these shoes don't fit?",
            "context": null,
            "trace_id": "eb1aba06-b71c-4ec9-98a2-11633f02337d",
            "timestamp": "20250225_145154",
            "example_id": "8602e43e-5f4d-4788-afed-224d4d58cef8",
            "tools_called": null,
            "actual_output": "We offer a 30-day full refund at no extra cost.",
            "example_index": null,
            "expected_tools": [
              "refund"
            ],
            "expected_output": "We offer a 30-day full refund at no extra cost.",
            "retrieval_context": [
              "All customers are eligible for a 30 day full refund at no extra cost."
            ],
            "additional_metadata": null
          }
       ]
 ]

### DIFF
--- a/src/judgeval/data/example.py
+++ b/src/judgeval/data/example.py
@@ -3,10 +3,11 @@ Classes for representing examples in a dataset.
 """
 
 
-from typing import TypeVar, Optional, Any, Dict, List
-from pydantic import BaseModel
+from typing import TypeVar, Optional, Any, Dict, List, UUID
+from pydantic import BaseModel, Field
 from enum import Enum
 from datetime import datetime
+from uuid import uuid4
 
 
 Input = TypeVar('Input')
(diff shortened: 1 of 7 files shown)


JCamyre (changes requested):
LGTM. To confirm, does this work with:
image

src/judgeval/data/example.py
```diff
@@ -33,15 +35,19 @@ class Example(BaseModel):
     tools_called: Optional[List[str]] = None
     expected_tools: Optional[List[str]] = None
     name: Optional[str] = None
-    example_id: Optional[str] = None
+    example_id: str = Field(default_factory=lambda: str(uuid4()))
+    example_index: Optional[int] = None
```
SecroLoL, line 39: Seems like `example_index` is replacing what used to be `example_id`. Is this correct?

Mandolaro: This is correct, and is intended

SecroLoL: Perfect

SecroLoL (approved):
LGTM, changed test cases to make them not fail

SecroLoL (commented):
Left changes to fix UTs

PR #84, opened 2025-02-26
Refactor Tracer

SecroLoL:
This PR updates the tracer syntax to be more uniform. 

1. Before, we had to write both `@observe` statements as well as the context manager `with judgment.trace(...) as trace`. Logically, this was a bit confusing and unnecessary. 

Now, we can just label every function that we want to trace with `@judgment.observe`! Easy as pie.

Before (main func):
```
async def get_food_recommendations(cuisine: str) -> str:
    """Main function to get restaurant recommendations."""
    # Search for restaurants
    with judgment.trace(name="...", project_name="...", overwrite=True):
      restaurants = await search_restaurants(cuisine)
      
      # Get menu highlights for each restaurant
      menu_items = {}
      for restaurant in restaurants:
          menu_items[restaurant['name']] = await get_menu_highlights(restaurant['name'])
          
      # Generate final recommendation
      recommendation = await generate_recommendation(cuisine, restaurants, menu_items)
```

After (main func):
```
@judgment.observe(span_type="Main Func")
async def get_food_recommendations(cuisine: str) -> str:
    """Main function to get restaurant recommendations."""
    # Search for restaurants
    restaurants = await search_restaurants(cuisine)
    
    # Get menu highlights for each restaurant
    menu_items = {}
    for restaurant in restaurants:
        menu_items[restaurant['name']] = await get_menu_highlights(restaurant['name'])
        
    # Generate final recommendation
    recommendation = await generate_recommendation(cuisine, restaurants, menu_items)
```

2. Project management in tracer init

We now define the project that the tracer routes to in the init:

`judgment = Tracer(project_name="my_project")`

and now, the trace names are no longer defined by the `name` field in the context manager; instead, it is simply a UUID4.

### DIFF
--- a/src/demo/cookbooks/new_bot/basic_bot.py
+++ b/src/demo/cookbooks/new_bot/basic_bot.py
@@ -1,0 +1,106 @@
+import os
+import asyncio
+from typing import Dict, List
+from openai import OpenAI
+from uuid import uuid4
+from dotenv import load_dotenv
+
+from judgeval.tracer import Tracer, wrap
+from judgeval.scorers import AnswerRelevancyScorer, FaithfulnessScorer
+
+# Initialize clients
+load_dotenv()
+judgment = Tracer(api_key=os.getenv("JUDGMENT_API_KEY"), project_name="restaurant_bot")
+client = wrap(OpenAI())
+
+@judgment.observe(span_type="Research")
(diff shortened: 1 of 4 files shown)


SecroLoL (commented):
Run UTs

PR #88, opened 2025-02-27
Update JudgmentClient with Dataset Editing Permissions

SecroLoL:
Adds functionality for client to edit datasets by adding `Example` and `GroundTruthExample`s to the dataset

also adds unit test for this feature


SecroLoL (commented):
Run UTs

PR #100, opened 2025-03-04
Fix LLM wrapping tracer call

SecroLoL:
Previously if you created the tracer after the OpenAI client, the tracer wouldn't work. That's because of the way we picked up the proper tracing client.

Now the client is picked up when the wrapped `client.chat.completions.create` call is made!

So now it doesn't matter whether you initialize one or the other first.

```
openai_client = wrap(OpenAI())
tracer = Tracer(project_name="mist-elasticsearch")
judgment_client = JudgmentClient()
```

or 
```
tracer = Tracer(project_name="mist-elasticsearch")
judgment_client = JudgmentClient()
openai_client = wrap(OpenAI())
```

### DIFF
--- a/src/judgeval/common/tracer.py
+++ b/src/judgeval/common/tracer.py
@@ -816,14 +816,15 @@ def wrap(client: Any) -> Any:
     Wraps an API client to add tracing capabilities.
     Supports OpenAI, Together, and Anthropic clients.
     """
-    tracer = Tracer._instance  # Get the global tracer instance
-    
     # Get the appropriate configuration for this client type
     span_name, original_create = _get_client_config(client)
     
     def traced_create(*args, **kwargs):
-        # Skip tracing if no active trace
-        if not (tracer and tracer._current_trace):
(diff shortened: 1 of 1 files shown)


src/judgeval/common/tracer.py
```diff
@@ -816,14 +816,15 @@ def wrap(client: Any) -> Any:
...
-    tracer = Tracer._instance  # Get the global tracer instance
-    
     # Get the appropriate configuration for this client type
     span_name, original_create = _get_client_config(client)
     
     def traced_create(*args, **kwargs):
-        # Skip tracing if no active trace
-        if not (tracer and tracer._current_trace):
+        # Get the current tracer instance (might be created after client was wrapped)
+        tracer = Tracer._instance
```
JCamyre, line 824: Does this fix work because we fetch the Tracer._instance when we call `traced_create()` rather than outside the function (which is called when we run `wrap()`, which takes place before we call `Tracer()`)?

JCamyre (approved):
One clarifying question, LGTM

PR #102, opened 2025-03-04
Type warning fix

Mandolaro:
Just fix the:
UserWarning: Pydantic serializer warnings:
Expected enum but got str with value 'answer_relevancy' - serialized value may not be as expected
Expected enum but got str with value 'faithfulness' - serialized value may not be as expected
Expected enum but got str with value 'faithfulness' - serialized value may not be as expected
Expected enum but got str with value 'answer_relevancy' - serialized value may not be as expected
Expected enum but got str with value 'answer_correctness' - serialized value may not be as expected
Expected enum but got str with value 'faithfulness' - serialized value may not be as expected
Expected enum but got str with value 'answer_relevancy' - serialized value may not be as expected
Expected enum but got str with value 'answer_correctness' - serialized value may not be as expected

By changing the validator


JCamyre (approved):
LGTM

PR #103, opened 2025-03-04
Update Judgee Tracking E2E Test

yifan1207:
- Changed test file to pytest
- Added organization ID to user headers to resolve 422 errors
- Improved Test Assertions: Modified count verification to check for increments rather than exact values
- Created simplified test case that avoids 500 errors
- Updated User/Org Tests: Adjusted expectations to reflect shared organization tracking

UPDATE (3/4 4:04 PM):

- Created and updated E2E tests to tests traces tracking and added edge case testing for judgee tracking


src/e2etests/test_judgee_update.py
```diff
@@ -1,598 +1,756 @@
 """
```
JCamyre, line 1: May need to update tests depending on the implementation changes you make on `judgment` side.

PR #105, opened 2025-03-05
Make the default trace name as the root function's name

SecroLoL:
title

Trace names in the UI will now default appear as the name of the root function when using the `observe` decorator.


SecroLoL (commented):
Rerun UTs

JCamyre (approved):
LGTM

PR #107, opened 2025-03-05
Edit Example Class 

SecroLoL:
changes the type of the `input` and `actual_output` vars of the `Example` class to be `str` rather than `TypeVar`.

also added `additional_metadata` as part of the `ExampleParams` enum


alanzhang25 (commented):
FAILED e2etests/test_rules_alerts.py::test_complex_rules - Exception: An unexpected error occurred during evaluation: An error occurred while executing the Judgment API request: Error...

Getting this error when I ran e2etests on your branch, but also getting the same error on main so don't think this is your doing. Just wanted to note

src/e2etests/test_datasets.py
alanzhang25, line 1: do we need this file? shouldnt the fixtures be accessible from the test_dataset_operations file as well

SecroLoL: Yeah you're right, I think this was a cursor thing lol

PR #113, opened 2025-03-10
Add Cornerstone Demo + Groundedness Scorer

SecroLoL:
Introduces some documents related to the Cornerstone demo (not important)

IMPORTANT CHANGES:
- Added the GroundednessScorer (only available through API on purpose).
    - The GroundednessScorer catches both factual inaccuracies between the output and retrieval context, as well as instruction-following mistakes
    
- Added Osiris Model Support for the scorer


SecroLoL (commented):
LGTM

PR #123, opened 2025-03-15
Handle multiple inputs

alanzhang25:
NOT READY FOR REVIEW YET

### DIFF
--- a/src/judgeval/data/example.py
+++ b/src/judgeval/data/example.py
@@ -37,6 +37,7 @@ class Example(BaseModel):
     example_index: Optional[int] = None
     timestamp: Optional[str] = None
     trace_id: Optional[str] = None
+    follow_up_inputs: Optional[Dict[str, str]] = None #For langgraph it is node/tool -> input
     
     def __init__(self, **data):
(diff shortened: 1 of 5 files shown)


src/judgeval/data/example.py
```diff
@@ -37,6 +37,7 @@ class Example(BaseModel):
     example_index: Optional[int] = None
     timestamp: Optional[str] = None
     trace_id: Optional[str] = None
+    follow_up_inputs: Optional[Dict[str, str]] = None #For langgraph it is node/tool -> input
```
SecroLoL, line 40: Seems like there could be a better way to abstract this. Perhaps the right abstraction is to have a datatype that subsumes the `Example` class and contains multiple `Example`s to simulate a workflow run's multiple steps?

SecroLoL (changes requested):
Not sure that this `followup_inputs` design is the right way to go. We can definitely discuss but I feel like there is a cleaner abstraction for chaining multiple steps together, such as having a datatype like `Workflow` that contains multiple `Example`s instead of adding multiple addtl inputs to `Example`

PR #125, opened 2025-03-17
Spinner for running Evaluation and Logging Results

alanzhang25:
https://linear.app/judgment/issue/JUD-398/loading-spinner-when-running-evaluations-with-use-judgment=true


SecroLoL (approved):
LGTM! Thank u

PR #128, opened 2025-03-19
Enable appending to an eval run

jack-devhub:
#### **Changes**
1. **Fixed `override` Parameter in `EvaluationRun` Initialization**:
   - Added the `override` parameter to the `EvaluationRun` object initialization in the `run_evaluation` method to ensure it is properly passed through the pipeline.

2. **Updated API Request Payload**:
   - Ensured the `override` parameter is included in the API request payload when logging evaluation results.

#### **Code Changes**
```python:src/judgeval/judgment_client.py
// ... existing code ...
eval = EvaluationRun(
    log_results=log_results,
    project_name=project_name,
    eval_name=eval_run_name,
    examples=examples,
    scorers=loaded_scorers,
    model=model,
    aggregator=aggregator,
    metadata=metadata,
    judgment_api_key=self.judgment_api_key,
    rules=loaded_rules,
    organization_id=self.organization_id,
    override=override  // Added this line
)
// ... existing code ...
```

```python:src/judgeval/run_evaluation.py
// ... existing code ...
response = requests.post(
    f"{ROOT_API}/log-evaluation-results/",
    headers={
        "Content-Type": "application/json",
        "Authorization": f"Bearer {evaluation_run.judgment_api_key}",
        "X-Organization-Id": evaluation_run.organization_id
    },
    json={
        "results": [result.to_dict() for result in merged_results],
        "project_name": evaluation_run.project_name,
        "eval_name": evaluation_run.eval_name,
        "override": evaluation_run.override  // Added this line
    },
    verify=True
)
// ... existing code ...
```

#### **Testing**
- **Test Case**: `test_override_eval` in `src/e2etests/test_eval_operations.py`
  - Verified that the `override` parameter is correctly passed through the pipeline.
  - Confirmed that the third run with `override=True` succeeds and the final non-override run fails as expected


alanzhang25 (changes requested):
I think we should have a e2e test for this, could you add one?

Also make sure that all current tests are passing for this.

saiman7:
> I think we should have a e2e test for this, could you add one?
> 
> Also make sure that all current tests are passing for this.

Addressed

PR #133, opened 2025-03-21
Add langchain dependencies from integration to pypoject.toml

SecroLoL:
previously, people had a bit of trouble with importing from `judgeval` since we upgraded the package to require the dependencies from our `langgraph` integration. Now, those are added to the package dependencies by default

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,12 @@ dependencies = [
     "anthropic",
     "nest-asyncio",
     "pika",
+    "openpyxl",
+    "langchain",
+    "langchain-huggingface",
+    "langchain-openai",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -28,15 +28,19 @@ dependencies = [
...
+    "langchain-core",
 ]
 
 [project.optional-dependencies]
 dev = [
     "pytest>=8.3.4",
     "pytest-asyncio>=0.25.0",
     "pytest-mock>=3.14.0",
-    "langfuse==2.50.3",
-    "patronus",
```
Mandolaro, line 39: I wonder why are we removing these dependencies.

PR #137, opened 2025-03-23
Update docs + error message for _success_check() method in Custom Scorers

SecroLoL:
https://linear.app/judgment/issue/JUD-538/bugfix-update-success-check-method-documentation

Previously, we had all scorers implement the success_check() method but now we have the _success_check() method. Update all error messages and documentation related to this!


alanzhang25 (commented):
LGTM! Just wondering why we increment package by .3

PR #144, opened 2025-03-28
add_from_csv refactor

brentju:
Previously, users had to supply all headers for the Example class when loading their datasets, even if not all fields would be used.

This commit gives users the flexibility to format their CSV files with custom labelled headers, provided the user passes in a dictionary mapping Judgment Example header -> CSV headers as a parameter. Custom delimiters for parsing fields such as `context`, `retrieval_context`, `tools`, etc. which accept lists can also be specified using the new `secondary_delimiter` parameter.

TESTING:
All tests are included in tests/data/datasets/test_dataset.py. The typecasting of the `additional_metadata` field into a dict is implicitly tested by the suite.


SecroLoL (approved):
LGTM! Thanks Brent

PR #146, opened 2025-03-28
Made sure proper UTC time passed to Supabase

adivate2021:
Fixed an issue where the times being passed to Supabase were not correct. They were stored as the local time of the user but as UTC, which is incorrect. Now they are the proper UTC time, and this allows the times to be displayed properly in the website.


adivate2021:
Preferrably someone who is not on Pacific Time takes this PR to ensure it works for all time zones

PR #149, opened 2025-03-30
Added Docs for Notification

Mandolaro:


docs/alerts/notifications.mdx
```diff
@@ -0,0 +1,158 @@
...
+Notifications are configured using the `NotificationConfig` class from the `judgeval.rules` module.
+
+### Configuration Options
+
+| Parameter | Type | Description |
+|-----------|------|-------------|
+| `enabled` | boolean | Whether notifications are enabled (default: `True`) |
+| `communication_methods` | list of strings | The methods to use for sending notifications (e.g., `["email", "slack"]`) |
+| `email_addresses` | list of strings | Email addresses to send notifications to |
+| `send_at` | integer (Unix timestamp) | Schedule notifications for a specific time |
```
SecroLoL, line 35: unclear what the send_at does when its not `None`. How do we set up recurring messages/notifs like every Friday or something?

docs/alerts/notifications.mdx
```diff
@@ -0,0 +1,158 @@
...
+
+### Basic Configuration
```
SecroLoL, line 37: Can you add another config example with `send_at` not set to `None`? What would that do

docs/alerts/notifications.mdx
```diff
@@ -0,0 +1,158 @@
...
+        Condition(metric=FaithfulnessScorer(threshold=0.7))
+    ],
+    combine_type="all",  # Trigger when all conditions fail
+    notification=notification_config
+)
+```
+
+## Scheduled Notifications
+
+You can schedule notifications to be sent at a specific time using the `send_at` parameter:
```
SecroLoL, line 92: Ah here it is lol

SecroLoL: Can you add a link in the upper docs area to here saying like (learn more about the `send_at` parameter here)

docs/alerts/notifications.mdx
```diff
@@ -0,0 +1,158 @@
...
+
+```python
+from judgeval.rules import NotificationConfig
+import time
+
+# Schedule notification for 1 hour from now
+one_hour_from_now = int(time.time()) + 3600
+
+notification_config = NotificationConfig(
```
SecroLoL, line 101: I don't rly understand how Notif config works for recurring notifications. Is it possible for us to group the notifications at some given time? So for example, we could just release the notifs in a report every day/hour?

This seems like for every rule, we have to send it 1 hour ahead. But this doesn't really do anything for people if its not aggregated into a report right? Otherwise, we'll have still a bunch of notifs but just delayed by an hour

docs/alerts/notifications.mdx
```diff
@@ -0,0 +1,158 @@
...
+rules = [
+    Rule(
+        name="Quality Check",
+        description="Check if all quality metrics meet thresholds",
+        conditions=[
+            # Only built-in APIScorers can be used as metrics
+            Condition(metric=FaithfulnessScorer(threshold=0.7)),
+            Condition(metric=AnswerRelevancyScorer(threshold=0.8))
+        ],
+        combine_type="all",  # Trigger when all conditions fail
```
SecroLoL, line 142: There should be an explanation of the various values for `combine_type`. Like `"any"` and `"all"`.

docs/alerts/platform_notifications.mdx
```diff
@@ -0,0 +1,74 @@
...
+| Recipient Emails | Email addresses that will receive the report |
+| Frequency | How often the report should be sent (Daily, Weekly, Monthly) |
+| Select Days | For weekly reports, specify which days of the week |
+| Time | When the report should be sent |
+| Timezone | Your local timezone for accurate scheduling |
+| Compare to Previous Period | Enable to see performance changes over time |
+
+Your reports will be sent automatically based on your schedule settings, providing insights into your model's performance over time.
+
+## Slack Integration
```
SecroLoL, line 47: Can you switch the order of this with the email one? People will likely be more interested in Slack than email

docs/alerts/rules.mdx
```diff
@@ -0,0 +1,109 @@
...
+
+### How Conditions are Evaluated
+
+When a condition is evaluated, it uses the scorer's threshold and internal evaluation logic:
+
+1. By default, a condition passes when the actual score is greater than or equal to the threshold
+2. If the scorer has a custom `success_check()` method, that method will be used instead
+3. The threshold is retrieved from the scorer's `threshold` attribute
+
+## Combine Types
```
SecroLoL, line 72: Ah, this is where it went. I highly recommend you add hyperlinks throughout the docs to the other keywords that are new to people like `condition`, `rule`, and `combine type`, etc.

SecroLoL (changes requested):
Looks good for the most part! Just a few comments to check out. Mainly, try hyperlinking more so that people aren't confused by new terms/syntax that we've invented such as `condition`, `rule`, and `combine type`

docs/alerts/notifications.mdx
Mandolaro: Unfortunately I have not done recurring notifications. Main reason is because this was built for offline evals, and not online evals, so I actually didn't see why we need to do it recurring. I think I will make a ticket to build this next PR.

docs/alerts/platform_notifications.mdx
Mandolaro: resolved

docs/alerts/notifications.mdx
Mandolaro: resolved

SecroLoL (approved):
changes look good, thank you!

PR #152, opened 2025-04-01
Fix: multiple client initialization issue

jack-devhub:
## 🎯 Issue
Fixes [JUD-570](https://linear.app/judgment/issue/JUD-570/multiple-judgmentclient-initialized-statements)

## 🔍 Problem
The `JudgmentClient` was printing initialization messages multiple times when instantiated, even though it should behave as a singleton. This occurred because:
1. The welcome message was printed in `__init__` without checking if the instance was already initialized
2. No proper singleton pattern was implemented
3. Multiple parts of the codebase creating new client instances would trigger duplicate messages

## 💡 Solution
Implemented a proper singleton pattern and initialization check:
1. Added singleton implementation using `__new__` method
2. Added initialization state tracking
3. Enhanced the welcome message formatting for better UX
4. Early return in `__init__` if already initialized

### Changes
```python:src/judgeval/judgment_client.py
class JudgmentClient:
    _instance = None
    
    def __new__(cls, *args, **kwargs):
        if cls._instance is None:
            cls._instance = super().__new__(cls)
        return cls._instance

    def __init__(self, judgment_api_key: str = os.getenv("JUDGMENT_API_KEY"), organization_id: str = os.getenv("JUDGMENT_ORG_ID")):
        # Only initialize and print welcome message once
        if hasattr(self, 'initialized'):
            return
        
        # ... rest of initialization ...
        welcome_msg = f"Successfully initialized JudgmentClient, welcome back {response.get('detail', {}).get('user_name', 'user')}!"
        print(f"\n={'='*len(welcome_msg)}=\n{welcome_msg}\n={'='*len(welcome_msg)}=\n")
        
        self.initialized = True
```

### DIFF
--- a/src/judgeval/judgment_client.py
+++ b/src/judgeval/judgment_client.py
@@ -43,8 +43,16 @@ class DeleteEvalRunRequestBody(BaseModel):
     project_name: str
     judgment_api_key: str
 
+class SingletonMeta(type):
+    _instances = {}
 
-class JudgmentClient:
+    def __call__(cls, *args, **kwargs):
+        if cls not in cls._instances:
+            instance = super().__call__(*args, **kwargs)
+            cls._instances[cls] = instance
+        return cls._instances[cls]
+
+class JudgmentClient(metaclass=SingletonMeta):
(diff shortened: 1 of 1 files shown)


alanzhang25 (approved):
LGTM!

PR #165, opened 2025-04-08
Eval Results

alanzhang25:
ONLY MERGE WITH OTHER CORRESPONDINGS PRS:

https://judgmentlabs.slack.com/archives/C08L2UEQC0Y/p1744094986252399

### DIFF
--- a/src/judgeval/scorers/score.py
+++ b/src/judgeval/scorers/score.py
@@ -399,7 +399,6 @@ async def a_eval_examples_helper(
         scorer.error = None  # Reset scorer error
 
     # scoring the Example
-    scoring_result = generate_scoring_result(example)  # Creates ScoringResult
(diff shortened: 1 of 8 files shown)


src/judgeval/scorers/score.py
```diff
@@ -411,18 +410,20 @@ async def a_eval_examples_helper(
 
     # Now that all the scoring functions of each scorer have executed, we collect 
     # the results and update the ScoringResult with the scorer data
+    success = True
```
brentju, line 413: is this `success` variable used anywhere else?

brentju (approved):
lgtm

src/judgeval/scorers/score.py
alanzhang25: No we had it before but it wasn't being used, just keeping it incase we need in the future.

PR #166, opened 2025-04-09
update .async_evaluate to use Example object for consistency with .ru…

NoahIslam:
…n_evaluation
https://linear.app/judgment/issue/JUD-644/async-evaluate-to-take-in-example-objects


src/demo/travel_agent.py
```diff
@@ -9,6 +9,7 @@
 
 from judgeval.tracer import Tracer, wrap
 from judgeval.scorers import AnswerRelevancyScorer, FaithfulnessScorer
+from judgeval.examples import Example
```
alanzhang25, line 12: Not sure if this is the right data import, I checked out the branch and this file is yellow

src/demo/travel_agent.py
```diff
@@ -218,12 +222,15 @@ async def create_travel_plan(destination, start_date, end_date, research_data):
...
     ).choices[0].message.content
 
-    judgment.async_evaluate(
-        scorers=[FaithfulnessScorer(threshold=0.5)],
+    example = Example(
         input=prompt,
         actual_output=str(response),
         retrieval_context=[str(vector_db_context), str(research_data)],
-        model="gpt-4o",
+        model="gpt-4o"
```
alanzhang25, line 229: no model field on Example object?

src/demo/travel_agent.py
```diff
@@ -125,11 +126,14 @@ async def get_flights(destination):
     """Search for flights to the destination."""
     prompt = f"Flights to {destination} from major cities"
     flights_search = search_tavily(prompt)
-    judgment.async_evaluate(
-        scorers=[AnswerRelevancyScorer(threshold=0.5)],
+    example = Example(
         input=prompt,
         actual_output=str(flights_search["results"]),
-        model="gpt-4o",
+        model="gpt-4o"
```
alanzhang25, line 132: no model field on Example object?

src/judgeval/common/tracer.py
```diff
@@ -1159,7 +1141,13 @@ async def async_wrapper(*args, **kwargs):
                 # Get current trace from contextvars
                 current_trace = current_trace_var.get()
                 if current_trace and scorers:
-                    current_trace.async_evaluate(scorers=scorers, input=args, actual_output=kwargs, model=model, log_results=log_results)
+                    # Create an Example object from args and kwargs
+                    example = Example(
+                        input=str(args),
+                        actual_output=str(kwargs),
+                        trace_id=current_trace.trace_id
+                    )
```
alanzhang25, line 1149: Not related to this PR, but I wonder what the purpose of the score function is? Gonna ask in slack

alanzhang25 (changes requested):
Nice job, just some import stuff for the demo. Also maybe we could add some comments to the documentation to say "Create example object to pass to async_evaluate" or something like that

alanzhang25 (changes requested):
Thanks for updates! I just thought of something else too, can we add some Scorer to Example checking to for the fields of the Example object. We do something similar in run_eval where we log to the user that a Example object might be missing certain fields for a scorer, so we should do the same here.

src/judgeval/scorers/score.py
```diff
@@ -21,6 +21,23 @@
 from judgeval.common.logger import example_logging_context, debug, error, warning, info
 from judgeval.judges import JudgevalJudge
 
+def check_examples(examples: List[Example], scorers: List[JudgevalScorer]) -> None:
```
alanzhang25, line 24: I think this might be too granular in the scoring. This applies to run_eval scoring too but we already do checking on the example fields there, I think this check should be at a higher level?

alanzhang25 (changes requested):

alanzhang25 (commented):
UTs

PR #177, opened 2025-04-15
Updated tracer

NoahIslam:
@yifan1207 Is adding backwards compatibility to work with the old syntax, ie directly passing in `input`, `actual_output`, etc.


alanzhang25 (commented):
Run UTs

JCamyre (commented):
UT's

PR #178, opened 2025-04-15
Langgraph Reintegration

alanzhang25:


src/judgeval/common/tracer.py
```diff
@@ -920,6 +934,18 @@ def __init__(
                 "To use a different project name, ensure the first Tracer initialization uses the desired project name.",
                 RuntimeWarning
             )
+    
+    def set_current_trace(self, trace: TraceClient):
+        """
+        Set the current trace context in contextvars
+        """
+        current_trace_var.set(trace)
```
JCamyre, line 942: What is the `contextvars`?

JCamyre (approved):
LGTM! Great work

PR #181, opened 2025-04-18
Removed unused braintrust_langchain import

justinsheu:
Was causing e2e errors because braintrust_langchain is not in Pipfile. The import is also unused


alanzhang25 (approved):
Sorry forgot to remove this, LGTM! Tests are still failing for other reasons.

PR #186, opened 2025-04-20
Added support for Gemini

adivate2021:
- Added support for both asynchronous and synchronous clients of Gemini (so now they show up in traces).


justinsheu (approved):
lgtm

PR #202, opened 2025-04-24
Integrating Tracing with Judgment Scorers

galen-topper:
Integrations with Braintrust, Langfuse, Langsmith, Arize tracing + judgment evaluations


alanzhang25 (dismissed):
Since these are custom facing changes, I think we should move this to our cookbooks repo and not within judgeval, intergrations/demos within judgeval should be only for internal use by devs

JCamyre (approved):
LGTM

PR #205, opened 2025-04-25
Fix bugs with deep tracing

yifan1207:
- deep tracing with also tracing print functions


docs/monitoring/tracing.mdx
```diff
@@ -111,7 +126,6 @@ const observedMyTool = judgment.observe({ spanType: "tool" })(myTool);
     Common span types include `tool`, `function`, `retriever`, `database`, `web search`, etc.
 </Note>
 
-
 #### Putting it all Together
 Here's a complete example of using judgeval's tracing mechanisms:
```
adivate2021, line 130: Should there be a main function in this? It seems like it based on the picture, but it is not in the code.

Also since we have automatic deep tracing, I think for this "Putting it all together" example, we should only have one observe decorator

docs/monitoring/tracing.mdx
```diff
@@ -263,21 +285,289 @@ runMainWithEvalTrace();
...
+    const processed = processQuery(query);
+    return await agent2(processed);
+});
+```
+</CodeGroup>
+
+## Example: Music Recommendation Agent
 In this video, we'll walk through all of the topics covered in this guide by tracing over a simple OpenAI API-based music recommendation agent.
 
-<iframe
```
adivate2021, line 270: Is there a reason for this change with the iframe? Also, we may need to update the video since it doesn't use deep tracing, but this could also be something for the future potentially.

docs/monitoring/tracing.mdx
```diff
@@ -263,21 +285,289 @@ runMainWithEvalTrace();
...
+  width="560"
+  height="315"
+  src="https://www.youtube.com/embed/7g0fut06UxQ"
+  title="YouTube video player"
+  frameborder="0"
+  allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture"
+  allowfullscreen
 ></iframe>
 
+## Advanced: Deep Tracing and Agent-to-Agent Calls ##
```
adivate2021, line 425: Is this a repeated section?

adivate2021 (changes requested):
I think the docs could use some changes (I gave some specific examples). The code for tracer.py and the demo LGTM.

adivate2021 (approved):
LGTM

PR #208, opened 2025-04-27
adding langchain openai/anthropic tracking

galen-topper:
adding functionality for langchain openai and anthropic wrapping!


alanzhang25 (changes requested):
We already have a file for the callback handler: check integrations/langgraph.py. Also if this demo file is for customer use, please move to cookbooks too

PR #209, opened 2025-04-27
Streaming tracing

Mandolaro:
Enable tracing for stream since it doesnt work when streaming is enable right now for LLM calls


Mandolaro:
UT

src/judgeval/common/tracer.py
```diff
@@ -1326,100 +1354,157 @@ def async_evaluate(self, *args, **kwargs):
...
-            # Skip tracing if no active trace
-            if not current_trace:
-                return original_create(*args, **kwargs)
+        is_streaming = kwargs.get("stream", False)
+
+        with current_trace.span(span_name, span_type="llm") as span:
+            input_data = _format_input_data(client, **kwargs)
+            span.record_input(input_data)
+
+            if isinstance(client, (AsyncOpenAI, OpenAI)) and is_streaming:
```
yifan1207, line 1378: for this would it be safer if we did not directly change the kwargs? Maybe considering if the users need to use it for other calls

yifan1207 (commented):
lgtm! just wodnering For very long streaming sessions do we want to add a mechanism to limit the size of full_content or to chunk it, considering the possible memory issues?

you want to also update the docs?

Mandolaro:
The point your raise is interesting but I wonder like where should we do it, like we are suppose to store log everything right? Not sure I am understanding your point. Will update docs also

Mandolaro:
The point your raise is interesting but I wonder like where should we do it, like we are suppose to store log everything right? Not sure I am understanding your point.

src/judgeval/common/tracer.py
Mandolaro: hmm yea, think what we do instead is tell user to turn it on if they want token data, since it is off be default

Mandolaro: addressed

yifan1207:
> The point your raise is interesting but I wonder like where should we do it, like we are suppose to store log everything right? Not sure I am understanding your point. Will update docs also

Ohh so basically I was thinking of that if we use string concatination, since string in immutable then we are creating a new string everything so maybe we could use a list and join at the end?

Mandolaro:
Make sense tbh, i forget string in python sucks

yifan1207 (approved):

PR #210, opened 2025-04-27
Litellm integration

galen-topper:
liteLLM integration using direct patching


alanzhang25 (changes requested):
Same note as other PR:

We already have a file for the callback handler: check integrations/langgraph.py. Also if this demo file is for customer use, please move to cookbooks too

PR #211, opened 2025-04-27
OpenAI Responses API Support

Mandolaro:
Added support for the new OpenAI API (we were only tracking Chat completions instead of Responses before).


yifan1207 (commented):
lgtm

Mandolaro:
I think for now let's not do that since I think we don't know what users want just yet, but might be useful for sure.

yifan1207 (approved):

PR #212, opened 2025-04-28
Tracing doc update

galen-topper:
adding documentation for better tracing wrap, also cleaning up


alanzhang25 (changes requested):
When u make updates to ur other PRs, if its related enough, just add the documentation as its own commit to that PR, not worth as its own PR. Similiar requests with already existing callback Handler (there is documentation for this too)

PR #223, opened 2025-05-02
Revert "Integrating Tracing with Judgment Scorers"

JCamyre:
Reverts JudgmentLabs/judgeval#202

Need to wait on these docs until we clean up the integration code.
@alanzhang25 Is going to add a cleaner way to import `use_judgment_scorer` for each platform


joshle298 (approved):
revert_boat

PR #246, opened 2025-05-14
Removed trace entries

adivate2021:
## 📝 Summary

Removes trace entries and uses trace spans directly

## 🎯 Purpose

Makes the code easier to read without trace entries


justinsheu (approved):
lgtm

PR #248, opened 2025-05-15
Ahh/span error reliability

abhishekg999:
## ✅ Checklist

- [ ] Self-review
- [ ] Video demo of changes
- [ ] Reviewers assigned

### DIFF
--- a/src/demo/errors.py
+++ b/src/demo/errors.py
@@ -1,0 +1,34 @@
+from judgeval.tracer import Tracer
+
+judgment = Tracer(project_name="errors")
+
+@judgment.observe(span_type="func")
+def a(a):
+    return b(a)
+
+def b(b):
+    try:
+        return c(b)
+    except Exception as e:
+        raise Exception("Error in b")
+
+def c(c):
+    return d(c)
+
+def d(d):
+    raise Exception("Error in d")
+
+@judgment.observe(span_type="func")
+def e(e):
+    f(e)
+
+def f(f):
+    g(f)
+
+def g(g):
+    return "OK"
+
+
+
+a(1)
+e(1)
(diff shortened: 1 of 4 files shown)


src/demo/errors.py
```diff
@@ -0,0 +1,34 @@
+from judgeval.tracer import Tracer
```
JCamyre, line 1: Couple notes:
1. How are you storing the error results in the database? It'd be great to have a dedicated field for this so we can easily aggregate stats for dashboard purposes. Talk to @alanzhang25 about this!
2. More of a frontend design question - I want to brainstorm the best way to indicate an error occurred on the span/Trace itself.

src/judgeval/data/trace.py
```diff
@@ -1,18 +1,114 @@
...
+import json
+from datetime import datetime
 
 class TraceSpan(BaseModel):
     span_id: str
     trace_id: str
     function: Optional[str] = None
     depth: int
-    created_at: Optional[str] = None
+    created_at: Optional[float] = None
```
JCamyre, line 12: Does this work with our database/backend?

JCamyre (changes requested):
This is SICK! Couple of questions.

src/judgeval/data/trace.py
abhishekg999: this is from this pr: https://github.com/JudgmentLabs/judgeval/pull/246

src/judgeval/common/tracer.py
```diff
@@ -49,6 +50,7 @@
 
 # Local application/library-specific imports
 from judgeval.constants import (
+    JUDGMENT_TRACES_ADD_ANNOTATION_API_URL,
```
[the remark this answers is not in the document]
abhishekg999: same format as other API url constants

JCamyre (approved):
LGTM! Ik there are some backend changes to propagate regarding the error tracking, however I want Juniper to try this now.
The backend changes will be in a Judgment PR anyways so it's fine.

PR #251, opened 2025-05-16
Cleaned up some e2etests that are being skipped

adivate2021:
## 📝 Summary

Removed e2etests that were skipped

## 🎯 Purpose

These tests were either broken or needed to be refactored.


Mandolaro (approved):
LGTM

PR #256, opened 2025-05-18
Update README to add granular cost tracking per user and per task

SecroLoL:
## 📝 Summary

Updated README to add granular cost tracking per user and per task

## 🎯 Purpose

Adds more depth into the way we improve cost tracking compared to any other solution


justinsheu (approved):
lgtm

PR #261, opened 2025-05-19
Adding Customer id & tags

yifan1207:
## 📝 Summary

- Adding cusomter id and tags columns in DB and adding set_metadata() function to enable users to update and set metadata of a trace including
 Supported keys:
        - customer_id: ID of the customer using this trace
        - tags: List of tags for this trace
        - has_notification: Whether this trace has a notification
        - overwrite: Whether to overwrite existing traces
        - rules: Rules for this trace
        - name: Name of the trace
        
      
- ## 🎯 Purpose

- give user ability to track customer usage of agents
- give user ability to add tags to better carategotize and filter their traces

 - ## Merge with Judgment PR https://github.com/JudgmentLabs/judgment/pull/339
## ✅ Checklist

- [ ] Self-review
- [ ] Video demo of changes
- [ ] Reviewers assigned

### DIFF
--- a/src/demo/customer_id_and_tags.py
+++ b/src/demo/customer_id_and_tags.py
@@ -1,0 +1,199 @@
+#!/usr/bin/env python3
+"""
+Demo script for setting and verifying customer ID and tags in Judgment traces.
+
+This script demonstrates how to set metadata (customer ID and tags) on traces
+and verifies that the metadata is correctly set and saved.
+
+Usage:
+    python customer_id_and_tags.py [--api-url API_URL] [--project PROJECT_NAME]
+"""
+
+import os
+import sys
+import json
+import random
+import logging
+import argparse
+from typing import List, Dict, Any, Optional, Tuple
+from pathlib import Path
+
+# Setup logging
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s - %(name)s - %(levelname)s - %(message)s"
+)
+logger = logging.getLogger("metadata-demo")
+
+try:
+    from judgeval.common.tracer import Tracer
+    from dotenv import load_dotenv
+except ImportError as e:
+    logger.error(f"Required package not found: {e}")
+    logger.error("Please install required packages: pip install judgeval python-dotenv")
+    sys.exit(1)
+
+# Initialize a global variable for the Tracer instance
+judgment = None
+
+
+def parse_arguments() -> argparse.Namespace:
+    """Parse command line arguments."""
+    parser = argparse.ArgumentParser(description="Test customer ID and tags in Judgment traces")
+    parser.add_argument(
+        "--api-url", 
+        default=None,
+        help="Judgment API URL (default: uses .env or http://localhost:8000)"
+    )
+    parser.add_argument(
+        "--project", 
+        default="metadata-test",
+        help="Project name for the trace (default: metadata-test)"
+    )
+    return parser.parse_args()
+
+
+def validate_environment() -> Tuple[Optional[str], Optional[str]]:
+    """Validate that required environment variables are set."""
+    # Load environment variables from .env file
+    load_dotenv()
+    
+    api_key = os.getenv("JUDGMENT_API_KEY")
+    org_id = os.getenv("JUDGMENT_ORG_ID")
+    
+    if not api_key:
+        logger.error("JUDGMENT_API_KEY environment variable is not set")
+        return None, org_id
+        
+    if not org_id:
+        logger.error("JUDGMENT_ORG_ID environment variable is not set")
+        return api_key, None
+        
+    return api_key, org_id
+
+
+def get_customer_id() -> str:
+    """Generate a random customer ID for testing."""
+    return f"customer-{random.randint(1000, 9999)}"
+
+
+def get_test_tags() -> List[str]:
+    """Generate a list of tags for testing."""
+    all_tags = ["test", "example", "metadata", "demo", "development"]
+    # Select 2-4 random tags
+    num_tags = random.randint(2, min(4, len(all_tags)))
+    return random.sample(all_tags, num_tags)
+
+
+def save_trace_data(trace_data: Dict[str, Any], filename: str = "trace_metadata.json") -> None:
+    """Save trace data to a JSON file."""
+    try:
+        with open(filename, "w") as f:
+            json.dump(trace_data, f, indent=2)
+        logger.info(f"Trace metadata saved to {filename}")
+    except Exception as e:
+        logger.error(f"Failed to save trace data to {filename}: {e}")
+
+
+def test_metadata(tracer: Tracer, customer_id: str, tags: List[str]) -> str:
+    """Test setting and verifying customer ID and tags."""
+    try:
+        logger.info(f"Setting customer ID: {customer_id}")
+        logger.info(f"Setting tags: {tags}")
+        
+        # Create a trace context
+        with tracer.trace(name="test_metadata") as trace:
+            # Set both customer ID and tags at once
+            tracer.set_metadata(
+                customer_id=customer_id,
+                tags=tags
+            )
+            
+            # Verify the metadata was set correctly
+            current_trace = tracer.get_current_trace()
+            if current_trace is None:
+                raise ValueError("Failed to get current trace")
+                
+            if current_trace.customer_id != customer_id:
+                logger.warning(f"Customer ID mismatch: expected '{customer_id}', got '{current_trace.customer_id}'")
+            else:
+                logger.info(f"Customer ID verified: {current_trace.customer_id}")
+                
+            if current_trace.tags != tags:
+                logger.warning(f"Tags mismatch: expected {tags}, got {current_trace.tags}")
+            else:
+                logger.info(f"Tags verified: {current_trace.tags}")
+            
+            # Print all metadata
+            logger.info("Printing trace metadata:")
+            tracer.print_metadata()
+            
+            # Do some work
+            logger.info("Performing test operations...")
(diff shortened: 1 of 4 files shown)


src/judgeval/common/tracer.py
```diff
@@ -326,7 +328,74 @@ def set_current_span(self, span: Any):
...
+                self.has_notification = bool(v)
+            elif k == "overwrite":
+                self.overwrite = bool(v)
+            elif k == "rules":
+                self.rules = v
+            elif k == "name":
+                self.name = v
+            else:
+                supported_keys = ["customer_id", "tags", "has_notification", "overwrite", "rules", "name"]
+                raise ValueError(f"Unsupported metadata key: {k}. Supported keys: {supported_keys}")
```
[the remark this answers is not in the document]
JCamyre: True

src/judgeval/common/tracer.py
```diff
@@ -326,7 +328,74 @@ def set_current_span(self, span: Any):
...
+                tags=["production", "api-v2"],
+                has_notification=True
+            )
+            ```
+        """
+        # Handle both calling styles
+        metadata = {}
+        if key is not None and value is not None:
+            metadata[key] = value
+        metadata.update(kwargs)
```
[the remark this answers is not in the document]
JCamyre: Fair

src/judgeval/common/tracer.py
```diff
@@ -661,7 +730,8 @@ def save(self, overwrite: bool = False) -> Tuple[str, dict]:
             "evaluation_runs": [run.model_dump() for run in self.evaluation_runs],
             "overwrite": overwrite,
             "parent_trace_id": self.parent_trace_id,
-            "parent_name": self.parent_name
+            "customer_id": self.customer_id,
+            "tags": self.tags
```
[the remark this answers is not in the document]
JCamyre: Yes, because we are factoring out sequence

trace_metadata.json
```diff
@@ -0,0 +1,11 @@
+{
```
JCamyre, line 1: Delete this file

JCamyre (changes requested):
Looks great Yifan! Couple small requests

src/demo/customer_id_and_tags.py
SecroLoL, line 1: I think we removed the `src/demo/` folder a few PRs ago.

I'm gonna move this script to the cookbooks repo!

SecroLoL: I also think we should remove this file in that case

SecroLoL: https://github.com/JudgmentLabs/judgment-cookbook/pull/12

src/judgeval/common/tracer.py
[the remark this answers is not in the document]
yifan1207: fixed!

src/judgeval/common/tracer.py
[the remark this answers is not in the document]
yifan1207: fixed! allowed userts to set to None

src/judgeval/common/tracer.py
```diff
@@ -304,6 +304,8 @@ def __init__(
         self.enable_evaluations = enable_evaluations
         self.parent_trace_id = parent_trace_id
         self.parent_name = parent_name
+        self.customer_id: Optional[str] = None  # Added customer_id attribute
+        self.tags: Optional[List[str]] = None  # Added tags attribute
```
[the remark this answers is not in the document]
yifan1207: added!

src/demo/customer_id_and_tags.py
yifan1207: okk!

src/judgeval/common/tracer.py
```diff
@@ -901,6 +908,107 @@ def save_with_rate_limiting(self, overwrite: bool = False, final_save: bool = Fa
...
+            key: The metadata key to set
+            value: The metadata value to set (can be None to unset/clear the value)
+            **kwargs: Additional metadata as keyword arguments
+        
+        Supported keys:
+        - customer_id: ID of the customer using this trace
+        - tags: List of tags for this trace
+        - has_notification: Whether this trace has a notification
+        - overwrite: Whether to overwrite existing traces
+        - rules: Rules for this trace
```
JCamyre: How does this field relate to Minh's rules?

src/judgeval/common/tracer.py
```diff
@@ -901,6 +908,107 @@ def save_with_rate_limiting(self, overwrite: bool = False, final_save: bool = Fa
     def delete(self):
         return self.trace_manager_client.delete_trace(self.trace_id)
     
+    def set_metadata(self, key: str = None, value: Any = None, **kwargs):
```
JCamyre, line 911: I feel having both key, val and kwargs as arguments is a little confusing.
Thoughts on just having `**kwargs`?

src/judgeval/common/tracer.py
```diff
@@ -901,6 +908,107 @@ def save_with_rate_limiting(self, overwrite: bool = False, final_save: bool = Fa
...
+    def set_tags(self, tags: List[str]):
+        """
+        Set the tags for this trace.
+        
+        Args:
+            tags: List of tags to set
+        """
+        self.set_metadata(tags=tags)
+
+    def print_metadata(self):
```
JCamyre: Thinking either making this a private method `_print_metadata` or deleting entirely.
Do you see an application for this outside of debugging?

src/judgeval/common/tracer.py
```diff
@@ -1193,6 +1308,49 @@ def fallback_encoder(obj):
         except Exception as e:
             warnings.warn(f"Failed to send evaluation runs batch: {e}")
     
+    def _send_traces_batch(self, traces: List[Dict[str, Any]]):
```
JCamyre: Confused here - isn't this from Minh's PR? Not sure why it is showing as new here.
Also, does it make sense to send these in parallel

JCamyre (changes requested):
Couple quick questions, then good to merge :)

src/judgeval/common/tracer.py
yifan1207: I was just thinking we could have more compatibility. like the default should be key and val but also maybe kwargs if they need

src/judgeval/common/tracer.py
yifan1207: ohh like allow users to edit their rule as part of metadata if they need. should we allow this?

src/judgeval/common/tracer.py
yifan1207: I think people might want to use this to see what it is, just like a method od convenience?

src/judgeval/common/tracer.py
yifan1207: ohh I think yeaa they are directly from minh's PR? I was making some edit to add compatibility to the semi saving trace span so that bscailly updating metadata would also use the queue to edit

src/judgeval/common/tracer.py
JCamyre: Hmm not sure I follow

src/judgeval/common/tracer.py
JCamyre: Not necessary!

src/judgeval/common/tracer.py
JCamyre: Is it so they can do `set_metadata(key=..., value=...)`

src/judgeval/common/tracer.py
JCamyre: I don't think so - when people are tracing their agents in production they don't want this extra info to clutter their logs.

src/judgeval/common/tracer.py
JCamyre: Interesting that it is showing that you added the entire function. Which file did you copy this function from?

src/judgeval/common/tracer.py
```diff
@@ -1233,6 +1391,23 @@ def queue_evaluation_run(self, evaluation_run: EvaluationRun, span_id: str, span
             }
             self._span_queue.put(eval_data)
     
+    def queue_trace(self, trace_data: Dict[str, Any]):
```
JCamyre: Same here

src/judgeval/common/tracer.py
```diff
@@ -2216,6 +2390,55 @@ def async_evaluate(self, *args, **kwargs):
...
+        Args:
+            tags: List of tags to set
+        """
+        current_trace = self.get_current_trace()
+        if current_trace:
+            current_trace.set_tags(tags)
+        else:
+            warnings.warn("No current trace found, cannot set tags")
+
+    def print_metadata(self):
```
JCamyre: Think we should delete

JCamyre (changes requested):
Couple last things!

src/judgeval/common/tracer.py
yifan1207: okkk!

yifan1207: fixed!

src/judgeval/common/tracer.py
yifan1207: yesss

src/judgeval/common/tracer.py
yifan1207: okok delted

src/judgeval/common/tracer.py
yifan1207: ohh my bad now I see it basically I can just use minh's queing functions deleted these!

src/judgeval/common/tracer.py
JCamyre: 👍

PR #270, opened 2025-05-21
Agent Names added

adivate2021:
Add agent names + add support for chaining tool evals + traces

### DIFF
--- a/src/judgeval/common/tracer.py
+++ b/src/judgeval/common/tracer.py
@@ -961,6 +961,7 @@ class Tracer:
             self.initialized: bool = True
             self.enable_monitoring: bool = enable_monitoring
             self.enable_evaluations: bool = enable_evaluations
+            self.class_identifiers: Dict[str, str] = {}  # Dictionary to store class identifiers
 
             # Initialize S3 storage if enabled
             self.use_s3 = use_s3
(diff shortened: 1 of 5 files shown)


src/judgeval/common/tracer.py
```diff
@@ -1118,6 +1146,20 @@ def observe(self, func=None, *, name=None, span_type: SpanType = "span", project
...
+                span_name = original_span_name
+
+                if args and hasattr(args[0], '__class__'):
+                    class_name = args[0].__class__.__name__
+                    if (class_name in self.class_identifiers):
+                        if hasattr(args[0], self.class_identifiers[class_name]):
+                            instance_name = getattr(args[0], self.class_identifiers[class_name])
+                            span_name = f"{instance_name}.{span_name}"
+                        else:
+                            raise Exception(f"Attribute {self.class_identifiers[class_name]} does not exist for {class_name}. Check your identify() decorator.")
```
alanzhang25, line 1161: can we somehow comvine this logic into a helper func at all for resuability? the reason I ask is because I wonder if this captures cases that uses the deep_profiler and not our observe statements, not sure if u accounted for that

alanzhang25 (changes requested):
Does this work with profiler?

src/judgeval/common/tracer.py
adivate2021: Added a helper func, also works for deep tracing now

src/judgeval/common/tracer.py
```diff
@@ -697,6 +697,14 @@ def _get_qual_name(self, frame) -> str:
...
         except Exception:
             return f"{module_name}.{func_name}"
+        # if args and hasattr(args[0], '__class__'):
+        #     class_name = args[0].__class__.__name__
+        #     if (class_name in self.class_identifiers):
+        #         if hasattr(args[0], self.class_identifiers[class_name]):
+        #             instance_name = getattr(args[0], self.class_identifiers[class_name])
+        #             span_name = f"{instance_name}.{span_name}"
+        #         else:
+        #             raise Exception(f"Attribute {self.class_identifiers[class_name]} does not exist for {class_name}. Check your identify() decorator.")
```
alanzhang25, line 707: can we delete this?

src/judgeval/data/tool.py
```diff
@@ -16,4 +17,10 @@ def validate_tool_name(cls, v):
     def validate_parameters(cls, v):
         if v is not None and not isinstance(v, dict):
             warnings.warn(f"Parameters should be a dictionary, got {type(v)}", UserWarning)
+        return v
+    
+    @field_validator('agent_name')
+    def validate_agent_name(cls, v):
+        if v is not None and not isinstance(v, str):
+            warnings.warn("Agent name should be a string, got {type(v)}", UserWarning)
```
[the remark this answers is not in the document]
alanzhang25: +1

alanzhang25 (changes requested):
small tweaks!

src/judgeval/common/tracer.py
adivate2021: Done

alanzhang25 (approved):
lgtm!

PR #286, opened 2025-05-28
modified deep tracing to chain with existing tracefuncs

justinsheu:


src/judgeval/common/tracer.py
```diff
@@ -739,7 +779,7 @@ def _trace(self, frame: types.FrameType, event: str, arg: Any):
                     break
             
             if not span_data:
-                return original_trace
+                return
```
Mandolaro, line 782: why are we not returning original_trace anymore

Mandolaro (changes requested):
seems fine, can you answer a clarification question

Mandolaro (approved):
LGTM

src/judgeval/common/tracer.py
justinsheu: addressed

PR #289, opened 2025-05-29
Ahh/http error recording

abhishekg999:
## 📝 Summary

Error recording for HTTP related errors.
These are now stored directly in JSON which makes status code and queriable from postgres

## ✅ Checklist

- [ ] Self-review
- [ ] Video demo of changes
- [ ] Reviewers assigned


src/judgeval/common/tracer.py
```diff
@@ -510,7 +504,15 @@ def record_usage(self, usage: TraceUsage):
             return span # Return the created entry
         # Removed else block - original didn't have one
         return None # Return None if no span_id found
-
+    
+    def record_error(self, error: Any):
```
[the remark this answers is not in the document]
yifan1207: do we want to standardize a model now of maybe in the future?

src/judgeval/common/tracer.py
```diff
@@ -562,7 +564,49 @@ def save(self, overwrite: bool = False) -> Tuple[str, dict]:
...
+    
+    # This is where we specially handle exceptions that we might want to collect additional data for.
+    # When we do this, always try checking the module from sys.modules instead of importing. This will
+    # Let us support a wider range of exceptions without needing to import them for all clients.
+    
+    # Most clients (requests, httpx, urllib) support the standard format of exposing error.request.url and error.response.status_code 
+    # The alternative is to hand select libraries we want from sys.modules and check for them:
+    # As an example:
+    
+    # try:
```
JCamyre: Why did you comment this out? We will try it now

JCamyre (changes requested):
Quick question about some commented-out code. LGTM besides that!

src/judgeval/common/tracer.py
abhishekg999: oh this is just a comment basically was just documenting how we might want to do it if we s want other specific libraries without actually importing them

JCamyre: 👍 
Can we support the common libraries natively and fall back to this generalized way?

JCamyre (approved):
LGTM!

PR #295, opened 2025-06-01
Agent state

andrewli2403:
## 📝 Summary

this PR allows for the checking of an agents internal state before a tool is called and after a tool is called. This occurs in the tracer object, and in the observe decorator definition to track when a new span is being created.

## 🎯 Purpose

this is neccessary for developers to understand the state of their agent before and after a tool call in case the tool is able to access the agents variables and what data the agent stores/

## 🎥 Demo of Changes

on slack

## 🧪 Testing

manually tested with mock-agent, wordproblemsolver

## ✅ Checklist

- [ X] Self-review
- [X ] Video demo of changes
- [X ] Unit Tests and CI/CD tests are passing
- [ X] Reviewers assigned

### DIFF
--- a/src/judgeval/common/tracer.py
+++ b/src/judgeval/common/tracer.py
@@ -1067,7 +1067,7 @@ class Tracer:
 
         rprint(f"[bold]{label}:[/bold] {msg}")
     
-    def identify(self, identifier: str, track_state: bool = False, track_attributes: Optional[List[str]] = None):
+    def identify(self, identifier: str, track_state: bool = False, track_attributes: Optional[List[str]] = None, field_mappings: Optional[Dict[str, str]] = None):
         """
         Class decorator that associates a class with a custom identifier.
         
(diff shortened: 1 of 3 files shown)


src/judgeval/common/tracer.py
```diff
@@ -1076,11 +1088,53 @@ class User:
...
+        """
+        Extract instance state if the instance should be tracked.
+        
+        Returns the captured state dict if tracking is enabled, None otherwise.
+        """
+        if args and hasattr(args[0], '__class__'):
+            instance = args[0]
+            class_name = instance.__class__.__name__
+            if (class_name in self.class_identifiers and 
+                isinstance(self.class_identifiers[class_name], dict) and
```
alanzhang25, line 1166: do we need to check if a dict?

alanzhang25 (approved):
one nit

PR #307, opened 2025-06-08
Langgraph integration traces can now track non-langgraph functions

justinsheu:
Non-langgraph functions (wrapped in .observe()) called from langgraph functions traced by Judgeval callback handler are now tracked within the same span.

For async langgraph workflows where non-langgraph functions may be traced in a different async context, trace_across_async_contexts must be set to True for the tracer

### DIFF
--- a/src/judgeval/common/tracer.py
+++ b/src/judgeval/common/tracer.py
@@ -333,15 +333,15 @@ class TraceClient:
 
     def get_current_span(self):
         """Get the current span from the context var"""
-        return current_span_var.get()
+        return self.tracer.get_current_span()
     
     def set_current_span(self, span: Any):
         """Set the current span from the context var"""
-        return current_span_var.set(span)
+        return self.tracer.set_current_span(span)
     
     def reset_current_span(self, token: Any):
         """Reset the current span from the context var"""
-        return current_span_var.reset(token)
+        self.tracer.reset_current_span(token)
         
     @contextmanager
     def span(self, name: str, span_type: SpanType = "span"):
@@ -351,8 +351,8 @@ class TraceClient:
         # Generate a unique ID for *this specific span invocation*
         span_id = str(uuid.uuid4())
         
-        parent_span_id = current_span_var.get() # Get ID of the parent span from context var
-        token = current_span_var.set(span_id) # Set *this* span's ID as the current one
+        parent_span_id = self.get_current_span() # Get ID of the parent span from context var
+        token = self.set_current_span(span_id) # Set *this* span's ID as the current one
         
         current_depth = 0
         if parent_span_id and parent_span_id in self._span_depths:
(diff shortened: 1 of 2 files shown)


src/judgeval/integrations/langgraph.py
```diff
@@ -142,6 +148,9 @@ def _end_span_tracking(
 
         # Get span ID and check if it exists
         span_id = self._run_id_to_span_id.get(run_id)
+        token = self.span_id_to_token.get(span_id)
+        if token:
+            current_span_var.reset(token)
```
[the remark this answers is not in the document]
justinsheu: resolved

src/judgeval/integrations/langgraph.py
```diff
@@ -167,6 +176,10 @@ def _end_span_tracking(
                 if self._trace_client and not self._trace_saved: # Check if not already saved
                     # TODO: Check if trace_client.save needs await if TraceClient becomes async
                     trace_id, trace_data = self._trace_client.save(overwrite=self._trace_client.overwrite) # Use client's overwrite setting
+                    token = self.trace_id_to_token.get(trace_id)
+                    if token:
+                        current_trace_var.reset(token)
```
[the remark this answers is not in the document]
justinsheu: resolved

src/judgeval/common/tracer.py
```diff
@@ -957,7 +970,8 @@ def __init__(
         s3_aws_secret_access_key: Optional[str] = None,
         s3_region_name: Optional[str] = None,
         offline_mode: bool = False,
-        deep_tracing: bool = True  # Deep tracing is enabled by default
+        deep_tracing: bool = True,  # Deep tracing is enabled by default
+        trace_across_async_contexts: bool = False # BY default, we don't trace across async contexts
```
Mandolaro, line 974: Can you explain what this change is for. Does it relate to how you are changing how we use contextvar

Mandolaro (changes requested):
Just a quick clarification otherwise look good

src/judgeval/common/tracer.py
justinsheu: Basically async langgraph workflow with ainvoke does tracing in a different async context then the tracing done through .observe() (for example if you look at the async demo I linked to in my PR slack post). This means that context variables cannot be shared between the two contexts and so I added a way to keep track of current trace and current span via the tracer object itself which is accessible across contexts. But you might still have some tracing workflows where each async context should be its own trace which is why I added this option

Also I mentioned this in a comment but instead of having member variable for Tracer instance keep track of current trace/span, I was thinking we could just use class/static variable since Tracer is a singleton anyways. But I'm gonna leave it like this for now in case we ever decide to make Tracer not a singleton

Mandolaro: Oh i remember this issue now. Sounds good

PR #309, opened 2025-06-09
Trace saving migration

Mandolaro:
## 📝 Summary

Migrating for saving full traces to saving spans:
1. Added background service that queue traces span
2. Added new functions that call new endpoint that perform spans saving
3. Fix assert test for langgraph
4. Modify view trace logic so now it show at start of execution

Merge this first: https://github.com/JudgmentLabs/judgment/pull/421
## 🎯 Purpose

Allow for live tracing + reduce amount of saving at once we do for a big trace by decomposing them into their span component

## 🧪 Testing

Pass test

## ✅ Checklist

- [ ] Self-review
- [ ] Video demo of changes
- [ ] Reviewers assigned

## 📌 Linear Issue

https://linear.app/judgment/issue/JUD-1071/migrate-trace-saving-from-saving-all-the-spans-to-saving-single-spans


src/judgeval/integrations/langgraph.py
```diff
@@ -137,6 +150,11 @@ def _start_span_tracking(
         new_trace.inputs = inputs
 
         trace_client.add_span(new_trace)
```
JCamyre: Thoughts on renaming `new_trace` -> `new_span`? Makes more sense to me

src/judgeval/integrations/langgraph.py
```diff
@@ -177,12 +200,33 @@ def _end_span_tracking(
...
 
                 if self._trace_client and not self._trace_saved: # Check if not already saved
-                    # TODO: Check if trace_client.save needs await if TraceClient becomes async
-                    trace_id, trace_data = self._trace_client.save(overwrite=self._trace_client.overwrite) # Use client's overwrite setting
+                    # Flush background spans before saving the final trace
+
+                    
+                    # NEW: Use save_with_rate_limiting with final_save=True for final save
+                    trace_id, trace_data = self._trace_client.save_with_rate_limiting(
+                        overwrite=self._trace_client.overwrite,
```
JCamyre: I remember us having a discussion about the `overwrite` field. Is this necessary?

src/judgeval/integrations/langgraph.py
```diff
@@ -177,12 +200,33 @@ def _end_span_tracking(
...
                     # current_trace_var.set(None)
-                    self.traces.append(trace_data) # Leaving this in for now but can probably be removed
+                    
+                    # Store complete trace data instead of server response
+                    complete_trace_data = {
+                        "trace_id": self._trace_client.trace_id,
+                        "name": self._trace_client.name,
+                        "created_at": datetime.utcfromtimestamp(self._trace_client.start_time).isoformat(),
+                        "duration": self._trace_client.get_duration(),
+                        "trace_spans": [span.model_dump() for span in self._trace_client.trace_spans],
```
JCamyre, line 210: Idea here is we rely on our own local storage of the trace_spans instead of asking the server for the spans?

src/judgeval/integrations/langgraph.py
```diff
@@ -177,12 +200,33 @@ def _end_span_tracking(
...
+                    trace_id, trace_data = self._trace_client.save_with_rate_limiting(
+                        overwrite=self._trace_client.overwrite, 
+                        final_save=True  # Final save with usage counter updates
+                    )
                     token = self.trace_id_to_token.pop(trace_id, None)
                     self.tracer.reset_current_trace(token, trace_id)
                     # current_trace_var.set(None)
-                    self.traces.append(trace_data) # Leaving this in for now but can probably be removed
+                    
+                    # Store complete trace data instead of server response
```
JCamyre: What does _end_span_tracking represent? Makes sense when you save complete trace for `on_chain_end`, not sure about `_end_span_tracking`

src/judgeval/run_evaluation.py
```diff
@@ -406,8 +406,15 @@ def run_trace_eval(trace_run: TraceRun, override: bool = False, ignore_errors: b
         )
     if function and tracer:
         new_traces: List[Trace] = []
-        tracer.offline_mode = True
-        tracer.traces = []
+        
+        # Handle case where tracer is actually a callback handler
+        actual_tracer = tracer
+        if hasattr(tracer, 'tracer') and hasattr(tracer.tracer, 'traces'):
+            # This is a callback handler, get the underlying tracer
```
JCamyre, line 413: This is to account for `JudgevalCallbackHandler`

src/tests/common/test_tracer.py
```diff
@@ -165,7 +165,10 @@ def test_save_trace(mock_post, trace_client):
     mock_response = Mock()
     mock_response.status_code = 200
     mock_response.text = '{"message": "success"}'
-    mock_response.json.return_value = {"ui_results_url": "http://example.com/results"}
+    mock_response.json.return_value = {
+        "ui_results_url": "http://example.com/results",
+        "trace_id": trace_client.trace_id
```
JCamyre, line 170: Where do we use the `trace_id` from `save()`?

src/judgeval/common/tracer.py
```diff
@@ -539,8 +740,12 @@ def record_output(self, output: Any):
             span = self.span_id_to_span[current_span_id]
             span.output = "<pending>" if inspect.iscoroutine(output) else output
             
-            if inspect.iscoroutine(output):
-                asyncio.create_task(self._update_coroutine(span, output, "output"))
+            # if inspect.iscoroutine(output):
```
JCamyre: Commented out because we are handling logic in `_update_coroutine`? If so, can we delete?

src/judgeval/common/tracer.py
```diff
@@ -603,14 +823,78 @@ def save(self, overwrite: bool = False) -> Tuple[str, dict]:
...
+            "evaluation_runs": [run.model_dump() for run in self.evaluation_runs],
+            "overwrite": overwrite,
+            "offline_mode": self.tracer.offline_mode,
+            "parent_trace_id": self.parent_trace_id,
+            "parent_name": self.parent_name
+        }
+        
+        # Check usage limits first
+        try:
+            usage_check_result = self.trace_manager_client.check_usage_limits(count=1)
```
JCamyre, line 869: Question: Does it make more sense to return True/False here? Try/excepts + Raise exceptions are slower than checking boolean values. We should keep exceptions to unexpected behavior. Thoughts?

src/judgeval/common/tracer.py
```diff
@@ -603,14 +823,78 @@ def save(self, overwrite: bool = False) -> Tuple[str, dict]:
...
+            try:
+                usage_update_result = self.trace_manager_client.update_usage_counters(count=1)
+                # Usage updated silently - no need to show detailed usage info
+            except ValueError as e:
+                # Log warning but don't fail the trace save since the trace was already saved
+                warnings.warn(f"Usage counter update failed (trace was still saved): {e}")
+
+        # Upload annotations
+        # TODO: batch to the log endpoint
+        for annotation in self.annotations:
```
JCamyre, line 895: Side note: Need to investigate if the backend implementation of annotations are worth saving. I don't trust it intuitively

src/judgeval/common/tracer.py
```diff
@@ -647,6 +931,326 @@ def _capture_exception_for_trace(current_trace: Optional['TraceClient'], exc_inf
...
+        except Exception as e:
+            warnings.warn(f"Failed to send evaluation runs batch: {e}")
+    
+    def queue_span(self, span: TraceSpan, span_state: str = "input"):
+        """
+        Queue a span for background sending.
+        
+        Args:
+            span: The TraceSpan object to queue
+            span_state: State of the span ("input", "output", "completed")
```
JCamyre, line 1190: Does "completed" refer to the final span in a Trace?

src/judgeval/common/tracer.py
```diff
@@ -1142,6 +1763,10 @@ def trace(
                 # Save the trace to the database to handle Evaluations' trace_id referential integrity
                 yield trace
             finally:
+                # Flush background spans for this trace before completing
```
JCamyre: Can we delete?

JCamyre (changes requested):
Minh this is so sick! Some questions to clarify then good to merge! Great stuff

src/judgeval/integrations/langgraph.py
Mandolaro: no it isnt anymore

src/judgeval/integrations/langgraph.py
Mandolaro: yea that is more reliable and is non blocking

src/judgeval/integrations/langgraph.py
Mandolaro: this basically just finishing span tracking

src/judgeval/run_evaluation.py
Mandolaro: this basically a bug where our for some reason when we do assert test the tracer is different from each other in assert test and langgraph

src/tests/common/test_tracer.py
Mandolaro: probably not necessary

src/judgeval/common/tracer.py
Mandolaro: oh accidentially comment this out

src/judgeval/common/tracer.py
Mandolaro: yea that is for later

src/judgeval/common/tracer.py
Mandolaro: yea should be final since it is when span context end

src/judgeval/common/tracer.py
Mandolaro: resolved

src/judgeval/integrations/langgraph.py
Mandolaro: blocking is always bad since our latency will always be off

src/judgeval/common/tracer.py
Mandolaro: yea later pr

src/judgeval/common/tracer.py
Mandolaro: that would require fixing all client side function since we are raising ValueError for all of them right now.

would be worth looking into in a future PR

src/judgeval/common/tracer.py
Mandolaro: resolve this issue

src/judgeval/integrations/langgraph.py
Mandolaro: resovled

src/judgeval/run_evaluation.py
JCamyre: 👍

src/judgeval/common/tracer.py
JCamyre: 👍

PR #320, opened 2025-06-17
Combine CI files + add branch check as conditional (JUD-1153)

justinsheu:
1. Branch protection workflow is now a dependency for tests to run
2. Combined branch protection workflow and the two ci workflows into one


Mandolaro (approved):
LGTM

PR #322, opened 2025-06-18
account for client.beta.chat.completions.parse

justinsheu:
both async and sync versions of client.beta.chat.completions.parse are now traced


JCamyre (approved):
LGTM!

PR #330, opened 2025-06-22
Judgment Datatypes (JUD-1212)

adivate2021:
- Use a script to get the important datatypes based on relevant routes
- These datatypes will be used instead of the current data directory being used
- Since OpenAPI cannot support methods for the Pydantic models, refactored them into new Pydantic models which extend the old ones
- Run transform.py to get the new datatypes
- Currently storing these datatypes in a file called judgment_types.py

Every time there has been an update to the datatypes, run the script ./update_types.sh to recreate the new openapi_new.json file and also regenerate the judgment_types.py file with the new Pydantic models. The datatypes in src/judgeval/data inherit from the generated types in judgment_types.py, and this is where instance methods can be added to the models.

Merge with https://github.com/JudgmentLabs/judgment/pull/481


src/judgeval/common/utils.py
```diff
@@ -40,18 +40,21 @@ class CustomModelParameters(pydantic.BaseModel):
     litellm_base_url: str
 
     @pydantic.field_validator("model_name")
```
justinsheu, line 42: remove field validators here?

src/judgeval/data/example.py
```diff
@@ -22,21 +20,7 @@ class ExampleParams(Enum):
...
-    name: Optional[str] = None
-    example_id: str = Field(default_factory=lambda: str(uuid4()))
-    example_index: Optional[int] = None
-    created_at: Optional[str] = None
-    trace_id: Optional[str] = None
-
+class Example(ExampleJudgmentType):
     def __init__(self, **data):
         if "example_id" not in data:
             data["example_id"] = str(uuid4())
```
justinsheu, line 26: verify that this should be kept here. it seems like before pushing to the database judgeval needs a way to reference a specific example by its example_id (e.g. for scoring) before pushing to the DB, so this is likely still needed. We should probably have a cleaner way of generating the example_id though

pyproject.toml
```diff
@@ -31,6 +31,8 @@ dependencies = [
     "google-genai",
     "boto3",
     "matplotlib>=3.10.3",
+    "openapi-python-client>=0.25.0",
```
justinsheu: remove

src/judgeval/data/openapi_new.json
```diff
@@ -0,0 +1,2892 @@
+{
```
justinsheu: remove in script + add to .gitignore

src/judgeval/generate_models.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
justinsheu: remove

src/judgeval/data/custom_example.py
```diff
@@ -1,19 +1,5 @@
...
-    retrieval_context: Optional[List[str]] = None
-    additional_metadata: Optional[Dict[str, Any]] = None
-    tools_called: Optional[List[str]] = None
-    expected_tools: Optional[List[str]] = None
-    name: Optional[str] = None
-    example_id: str = Field(default_factory=lambda: str(uuid4()))
-    example_index: Optional[int] = None
-    timestamp: Optional[str] = None
-    trace_id: Optional[str] = None
+class CustomExample(CustomExampleJudgmentType):
```
justinsheu, line 4: note for future: eventually get rid of these wrapper types since we likely won't be adding methods to types

justinsheu (changes requested):
left a few small comments

src/judgeval/data/tool.py
```diff
@@ -1,56 +1,16 @@
...
-        return v
+from pydantic import model_serializer
+from typing import Dict, Any
+from judgeval.data.judgment_types import ToolJudgmentType
+
+
+class Tool(ToolJudgmentType):
+    pass
+
+    @model_serializer
```
justinsheu: verify this is also in judgment. is this necessary?

src/judgeval/data/trace.py
```diff
@@ -1,42 +1,24 @@
...
-    annotation: Optional[List[Dict[str, Any]]] = None
-    evaluation_runs: Optional[List[EvaluationRun]] = []
-    expected_tools: Optional[List[Tool]] = None
-    additional_metadata: Optional[Dict[str, Any]] = None
-    has_evaluation: Optional[bool] = False
-    agent_name: Optional[str] = None
-    state_before: Optional[Dict[str, Any]] = None
-    state_after: Optional[Dict[str, Any]] = None
+class TraceSpan(TraceSpanJudgmentType):
+    def get_name(self):
```
justinsheu: is this also in judgment

src/tests/scorers/test_judgeval_scorer.py
```diff
@@ -117,19 +117,19 @@ async def test_a_score_example_implementation(self, basic_scorer):
         score = await basic_scorer.a_score_example({"test": "example"})
         assert score == 0.9
 
-    def test_success_check_implementation(self, basic_scorer):
+    def testsuccess_check_implementation(self, basic_scorer):
```
justinsheu: add underscore back

src/tests/scorers/test_score.py
```diff
@@ -865,24 +865,26 @@ async def test_a_eval_examples_helper_success(
...
+                verbose_logs=mock_scorer_2.verbose_logs,
+                additional_metadata=mock_scorer_2.additional_metadata,
+            )
+        ]
         mock_create_scorer_data.return_value = mock_scorer_data
 
         mock_scoring_result = ScoringResult(
             success=True,
-            scorers_data=[mock_scorer_data],
+            scorers_data=[mock_scorer_data[0]],
```
justinsheu: should this be scorers_data=mock_scorer_data?

src/judgeval/generate_models.sh
adivate2021: done

src/tests/scorers/test_judgeval_scorer.py
adivate2021: done

pyproject.toml
adivate2021: done

src/judgeval/data/openapi_new.json
adivate2021: done

src/judgeval/data/example.py
adivate2021: agreed, made a linear ticket

src/judgeval/data/tool.py
adivate2021: removed

src/judgeval/data/trace.py
adivate2021: removed

src/tests/scorers/test_score.py
adivate2021: I think this should be ok for now since before, we would append the result of create_scorer_data to a list, but now we extend it. So by extracting the 0th index and putting it in a new list, it is kind of mimicking that behavior the way the old test would just have the scorer data within the list element.

src/judgeval/common/utils.py
adivate2021: I will keep them for now since they seem to have some logic. Can probably have a new PR in the future for refactoring out the field validators.

src/judgeval/data/custom_example.py
adivate2021: Makes sense to remove them, it will take some refactoring in the future.

src/update_types.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/zsh
+
+# Make sure Judgment backend server is running on port 8000
+# This openapi_transform.py will get the relevant parts of the openapi.json file and save it to openapi_new.json
+uv run judgeval/data/openapi_transform.py > judgeval/data/openapi_new.json
+
+# Then, datamodel-codegen will generate the judgment_types.py file based on the schema in openapi_new.json.
+datamodel-codegen --input judgeval/data/openapi_new.json --output judgeval/data/judgment_types.py --use-annotated
```
justinsheu, line 8: should this also call fix_default_factory?

src/judgeval/data/judgment_types.py
```diff
@@ -0,0 +1,321 @@
...
+
+class TraceJudgmentType(BaseModel):
+    trace_id: Annotated[str, Field(title="Trace Id")]
+    name: Annotated[str, Field(title="Name")]
+    created_at: Annotated[str, Field(title="Created At")]
+    duration: Annotated[float, Field(title="Duration")]
+    trace_spans: Annotated[List[TraceSpanJudgmentType], Field(title="Trace Spans")]
+    overwrite: Annotated[Optional[bool], Field(title="Overwrite")] = False
+    offline_mode: Annotated[Optional[bool], Field(title="Offline Mode")] = False
+    rules: Annotated[Optional[Dict[str, Any]], Field(title="Rules")] = {}
```
justinsheu: i'm assuming these should be converted to default factory right? looks like the script just wasn't run

src/update_types.sh
adivate2021: should be good now

src/tests/scorers/test_custom_scorer.py
```diff
@@ -116,19 +116,19 @@ async def test_a_score_example_implementation(self, basic_scorer):
         score = await basic_scorer.a_score_example({"test": "example"})
         assert score == 0.9
 
-    def test_success_check_implementation(self, basic_scorer):
+    def testsuccess_check_implementation(self, basic_scorer):
```
justinsheu: change test name here?

adivate2021: done

justinsheu (approved):
lgtm

PR #332, opened 2025-06-23
Now account for client.beta.chat.completions.parse correctly for wrapped OpenAI clients

justinsheu:
Wrapped `client.beta.chat.completions.parse` was previously being treated as another instance of `client.chat.completions.create` although `client.beta.chat.completions.parse` takes in an extra "response_format" field. Now also account for this correctly in the trace span input by also recording the response format, and in the trace span output by outputting "parsed" instead of "content"


JCamyre (approved):
LGTM!

PR #333, opened 2025-06-23
Removed singleton pattern

justinsheu:
Tracer object is no longer a singleton. This allows you to define multiple Tracer instances (with different project names) across a single execution


JCamyre (approved):
LGTM!

PR #339, opened 2025-06-25
Requests now have exponential backoff for retries (JUD-1257)

justinsheu:
To account for failing requests to the backend, HTTP requests now retry with exponential backoff. The schedule is as follows:
1st try -> 1st retry: 0.5s
1st retry -> 2nd retry: 1s
2nd retry -> 3rd retry: 2s
Request fails after the 3rd retry with a RetryError
This schedule can be modified in the utils/requests.py file


Mandolaro (approved):
LGTM

PR #342, opened 2025-06-28
Update how users save metadata to traces (JUD-1293) 

JCamyre:
## 📝 Summary

Update how users save metadata to traces.
```
    def get_rewards() -> float:
        return 1.0

    judgment.update_metadata(
        metadata={"reward_score": get_rewards()}
    ) # Set the rewards function
```

Exported trace as `json`:
```
[
  {
    "trace_id": "3484ecbd-24df-4a9f-8865-c2db25aacfe1",
    "metadata": {
      "reward_score": 1.0
    },
    "trace_spans": {
         # tools calls stored here
    },
    ... other Trace fields here
   }
]
```

## 🎯 Purpose

Better dev ergonomics.

## ✅ Checklist

- [ ] Self-review
- [ ] Video demo of changes
- [ ] Reviewers assigned

### DIFF
--- a/src/judgeval/common/tracer.py
+++ b/src/judgeval/common/tracer.py
@@ -650,8 +650,11 @@ class TraceClient:
             span.inputs = inputs
 
             # Queue span with input data
-            if self.background_span_service:
-                self.background_span_service.queue_span(span, span_state="input")
+            try:
+                if self.background_span_service:
+                    self.background_span_service.queue_span(span, span_state="input")
+            except Exception as e:
+                warnings.warn(f"Failed to queue span with input data: {e}")
 
     def record_agent_name(self, agent_name: str):
         current_span_id = self.get_current_span()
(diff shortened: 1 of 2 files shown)


abhishekg999:
The union type of tags differs:

```py
List[Union[str, set, tuple]]
```

vs

```py
def set_tags(self, tags: Union[List[str], str, dict, set, tuple]):
```

Which should be supported here

src/judgeval/common/tracer.py
```diff
@@ -419,7 +419,8 @@ def __init__(
         self.parent_trace_id = parent_trace_id
         self.parent_name = parent_name
         self.customer_id: Optional[str] = None  # Added customer_id attribute
-        self.tags: Optional[List[str]] = None  # Added tags attribute
+        self.tags: List[Union[str, set, tuple]] = []  # Added tags attribute
```
abhishekg999, line 422: typing of self.tags here: 

        self.tags: List[Union[str, set, tuple]] = []  # Added tags attribute

differs from the type: tags: Union[List[str], str, dict, set, tuple]

In the `set_tags` methods

PR #344, opened 2025-06-28
Try/Except Cases (JUD-1281 and JUD-1285)

alanzhang25:
## ✅ Checklist

- [ ] Self-review
- [ ] Video demo of changes
- [ ] Reviewers assigned


Mandolaro (changes requested):
You should also try except in the init also since this is where we validate api key and we dont try except there

PR #345, opened 2025-07-01
Scorer Cleanup (JUD-1275)

alanzhang25:
## 📝 Summary

1. APIScorerConfig/ (judgeval) ScorerConfig (judgment) Datatype - this data abstraction allows for cleaner use when sending scorers between judgeval and judgment.
2.  Got rid of the inconsistencies for naming of the scores (the name property or name field) and is now a property of ScorerConfig and BaseScorer
3. . Either Deleted or commented out scorers that are not referenced on the docs currently. This can be a initiative since we might want to keep some of the logic for the scorers.
4. . Threshold is now optional and defaults to 0.5
5. . Strict mode actually works for backend scorers (I have a linear ticket follow up for this though since it is not working properly: https://linear.app/judgment/issue/JUD-1309/threshold-for-scoring-should-not-set-score-to-zero-but-set-success)

## ✅ Checklist

- [ ] Self-review
- [ ] Video demo of changes
- [ ] Reviewers assigned


src/judgeval/constants.py
```diff
@@ -7,31 +7,25 @@
...
+    FAITHFULNESS = "Faithfulness"
+    ANSWER_RELEVANCY = "Answer Relevancy"
+    ANSWER_CORRECTNESS = "Answer Correctness"
+    INSTRUCTION_ADHERENCE = "Instruction Adherence"
+    EXECUTION_ORDER = "Execution Order"
+    DERAILMENT = "Derailment"
+    TOOL_ORDER = "Tool Order"
+    CLASSIFIER = "Classifier"
+    TOOL_DEPENDENCY = "Tool Dependency"
+    CUSTOM = "Custom"
```
Mandolaro, line 28: are we thinking of doing custom scorer on the backend?

alanzhang25: Nope but can leave there for readability and also so that we save that value in DB

PR #355, opened 2025-07-05
Tracer Clean

alanzhang25:
## 📝 Summary

Removing rules from tracer and also not allowing users to set offline_mode (this should be a user set attribute)

## ✅ Checklist

- [ ] Video demo of changes
- [ ] Reviewers assigned

### DIFF
--- a/src/judgeval/common/tracer.py
+++ b/src/judgeval/common/tracer.py
@@ -63,7 +63,6 @@ from judgeval.constants import (
 )
 from judgeval.data import Example, Trace, TraceSpan, TraceUsage
 from judgeval.scorers import APIScorerConfig, BaseScorer
-from judgeval.rules import Rule
 from judgeval.evaluation_run import EvaluationRun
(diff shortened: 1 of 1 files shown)


src/judgeval/common/tracer.py
```diff
@@ -402,18 +401,16 @@ def __init__(
...
-        rules: Optional[List[Rule]] = None,
         enable_monitoring: bool = True,
         enable_evaluations: bool = True,
         parent_trace_id: Optional[str] = None,
         parent_name: Optional[str] = None,
     ):
         self.name = name
         self.trace_id = trace_id or str(uuid.uuid4())
-        self.project_name = project_name or str(uuid.uuid4())
+        self.project_name = project_name or "default_project"
```
Mandolaro, line 411: i remember there was some discussion as to why we make it an UUID, can you try asking around.

Mandolaro (changes requested):
I think looks fine, just 1 comment

PR #388, opened 2025-07-11
JUD-1400: Non Rich Print Compatibility

alanzhang25:
## ✅ Checklist

- [ ] Video demo of changes
- [ ] Reviewers assigned


src/judgeval/judgment_client.py
```diff
@@ -113,7 +112,7 @@ def a_run_evaluation(
             append=append,
             async_execution=True,
         )
-        assert not isinstance(result, (asyncio.Task, SpinnerWrappedTask))
```
abhishekg999, line 114: Why do we assert here? Can this assertion ever fail and will it break their code?

src/judgeval/run_evaluation.py
```diff
@@ -492,19 +494,18 @@ def run_trace_eval(
         for example in examples:
             if example.input:
                 if isinstance(example.input, str):
-                    run_with_spinner(
-                        "Running agent function: ", function, example.input
-                    )
+                    judgeval_logger.info("Running agent function ...")
```
abhishekg999: (minor, nitpick) this log is repeated 3 times can be moved outside the branch

src/judgeval/run_evaluation.py
```diff
@@ -1130,34 +1039,42 @@ def assert_test(scoring_results: List[ScoringResult]) -> None:
...
 
         # Print individual test cases
         for i, result in enumerate(scoring_results):
             test_num = i + 1
             if result.success:
-                rprint(f"[green]✓ Test {test_num}: PASSED[/green]")
+                judgeval_logger.info(f"[green]✓ Test {test_num}: PASSED[/green]")
             else:
-                rprint(f"[red]✗ Test {test_num}: FAILED[/red]")
+                judgeval_logger.info(f"[red]✗ Test {test_num}: FAILED[/red]")
```
abhishekg999: Should this be .info? Ususally loggers provide and interface to .info, .warning, and .error. Would .error be more appropriate here? Few other places in the codebase where possibly we can use the more semantic version of the logger function

src/judgeval/common/tracer.py
```diff
@@ -71,6 +73,7 @@
 import queue
 import atexit
 
+console = Console()
```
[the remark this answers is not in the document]
abhishekg999: @alanzhang25 prolly valid comment. If we are already exporting the judgment_logger in logging.py is it possible to just import that;.

src/judgeval/judgment_client.py
alanzhang25: updated so its just asserting that its a asyncio task

PR #466, opened 2025-07-22
MyPy fix (JUD-1537)

justinsheu:
## 📝 Summary

- [ ] 1. separate mypy workflow
- [ ] 2. fix some mypy errors
- [ ] 3. install mypy to dev env
- [ ] 4. run different mypy command to capture more errors

## ✅ Checklist

- [ ] Video demo of changes
- [ ] Reviewers assigned


src/judgeval/common/utils.py
```diff
@@ -167,7 +169,9 @@ def fetch_together_api_response(
 
 
 async def afetch_together_api_response(
-    model: str, messages: List[Mapping], response_format: pydantic.BaseModel = None
+    model: str,
+    messages: List[Dict],
+    response_format: pydantic.BaseModel | None = None,
```
abhishekg999: The `TypeA | TypeB` syntax is not supported pre python 3.10. Prolly fine? But if we want to maintain compatibility maybe we can convert the usages of `|` to `Union[TypeA, TypeB]`?

I believe the rest of our codebase hopefully shouldnt be tied to only >3.10.

Thoughts?

abhishekg999 (changes requested):
I minor comment here regarding syntax compatibility.

src/judgeval/common/utils.py
justinsheu: that makes sense, i'll go through an change those instances

abhishekg999 (approved):
looks good

PR #475, opened 2025-07-25
 Example and Dataset Changes 

alanzhang25:
## 📝 Summary

- [ ] 1. Example is essential a dict that allows users to inherit from

- [ ] Exeriments can only be made up of the same example object keys! This applies to new ones and also when you append

- [ ] 2. Dataset is now a class that users do .get and .create from
- [ ] 3. Some more E2Etesting for new edge case scenarios

## ✅ Checklist

- [ ] Video demo of changes
- [ ] Reviewers assigned


src/judgeval/data/judgment_types.py
```diff
@@ -182,23 +166,15 @@ class TraceRunJudgmentType(BaseModel):
     traces: Annotated[List[TraceJudgmentType], Field(title="Traces")]
     scorers: Annotated[List[ScorerConfigJudgmentType], Field(title="Scorers")]
     model: Annotated[str, Field(title="Model")]
+    judgment_api_key: Annotated[Optional[str], Field(title="Judgment Api Key")] = None
```
Mandolaro: why is there judgment api key

Mandolaro (changes requested):
Just wondering if Example is just a legacy object now or people will still use it?

src/judgeval/data/example.py
```diff
@@ -15,47 +16,38 @@ class ExampleParams(str, Enum):
     RETRIEVAL_CONTEXT = "retrieval_context"
     TOOLS_CALLED = "tools_called"
     EXPECTED_TOOLS = "expected_tools"
-    REASONING = "reasoning"
     ADDITIONAL_METADATA = "additional_metadata"
 
 
-class Example(ExampleJudgmentType):
-    example_id: str = ""
+class Example(BaseModel):
```
Mandolaro: wait why is this not defined in Judgment?

alanzhang25: resolved

Mandolaro (approved):
LGTM

PR #488, opened 2025-07-31
Minor fixes to prompt scorer

adivate2021:
- Minor changes to prompt scorer to account for differences in backend
- Change set_name to check if the prompt scorer already exists

### DIFF
--- a/src/judgeval/scorers/judgeval_scorers/api_scorers/prompt_scorer.py
+++ b/src/judgeval/scorers/judgeval_scorers/api_scorers/prompt_scorer.py
@@ -32,7 +32,7 @@ def fetch_prompt_scorer(
 ):
     client = JudgmentApiClient(judgment_api_key, organization_id)
     try:
-        scorer_config = client.fetch_scorer(name)
+        scorer_config = client.fetch_scorer(name)["scorer"]
         scorer_config.pop("created_at")
         scorer_config.pop("updated_at")
         return scorer_config
@@ -121,8 +121,11 @@ class PromptScorer(APIScorerConfig):
         """
         Updates the name of the scorer.
         """
-        self.name = name
-        self.push_prompt_scorer()
+        if not scorer_exists(name, self.judgment_api_key, self.organization_id):
+            self.name = name
+            self.push_prompt_scorer()
+        else:
+            raise JudgmentAPIError(f"Scorer with name {name} already exists.")
 
     def set_threshold(self, threshold: float):
         """


src/judgeval/scorers/judgeval_scorers/api_scorers/prompt_scorer.py
```diff
@@ -1,21 +1,25 @@
...
 import os
 from judgeval.common.exceptions import JudgmentAPIError
+from copy import copy
 
 
 def push_prompt_scorer(
     name: str,
     prompt: str,
-    options: Mapping[str, float],
+    options: Optional[Mapping[str, float]] = None,
```
alanzhang25: is there a reason we use Mapping here instead of dict? so we dont have to cast later

src/judgeval/common/api/api.py
```diff
@@ -238,12 +238,18 @@ def check_example_keys(self, keys: List[str], eval_name: str, project_name: str)
...
+            payload: ScorerSavePayload = {
+                "name": name,
+                "prompt": prompt,
+                "options": options,
+            }
+        else:
+            payload: ScorerSavePayload = {
+                "name": name,
+                "prompt": prompt,
+            }
```
alanzhang25: can we just pass none options here?

src/judgeval/scorers/judgeval_scorers/api_scorers/prompt_scorer.py
```diff
@@ -121,8 +121,11 @@
         """
         Updates the name of the scorer.
         """
-        self.name = name
-        self.push_prompt_scorer()
+        if not scorer_exists(name, self.judgment_api_key, self.organization_id):
+            self.name = name
+            self.push_prompt_scorer()
+        else:
+            raise JudgmentAPIError(f"Scorer with name {name} already exists.")
```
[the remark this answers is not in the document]
alanzhang25: kind of +1, should we delete the old scorer

src/judgeval/scorers/judgeval_scorers/api_scorers/prompt_scorer.py
```diff
@@ -181,11 +188,17 @@ def get_config(self) -> dict:
...
+            return {
+                "name": self.name,
+                "prompt": self.prompt,
+                "options": self.options,
+            }
+        else:
+            return {
+                "name": self.name,
+                "prompt": self.prompt,
+            }
```
alanzhang25: think its fine to have Options be None here

src/judgeval/scorers/judgeval_scorers/api_scorers/prompt_scorer.py
adivate2021: Old code was using it, so I just kept it originally, but I agree with you. Just made the changes to all of the Mapping types to be Dict

src/judgeval/common/api/api.py
adivate2021: Updated

src/judgeval/scorers/judgeval_scorers/api_scorers/prompt_scorer.py
adivate2021: changed

src/judgeval/scorers/judgeval_scorers/api_scorers/prompt_scorer.py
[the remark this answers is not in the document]
adivate2021: just removed the set_name function

alanzhang25 (approved):
lgtm

PR #496, opened 2025-08-02
Local eval queue (JUD-1686)

Mandolaro:
## 📝 Summary

- [ ] 1. Added Local Evaluation Queue to enable async_evaluate for Local Custom Scorer

## ✅ Checklist

- [ ] Video demo of changes
- [ ] Reviewers assigned


src/judgeval/common/tracer/core.py
```diff
@@ -1612,9 +1628,68 @@ def shutdown_background_service(self):
...
+        Args:
+            timeout: Maximum time to wait in seconds. Defaults to 30 seconds.
+                    None means wait indefinitely.
+
+        Returns:
+            True if all processing completed within the timeout, False otherwise.
+
+        """
+        try:
+            judgeval_logger.info("Waiting for all evaluations and spans to complete...")
```
alanzhang25: can u ether delete all the loggers or change them to debug so it doesnt flood users terminal?

src/judgeval/__init__.py
alanzhang25, line 1: why do we add import here?

side note: we should clean the clients and also add Tracer here for easier import probably?

src/judgeval/common/tracer/core.py
```diff
@@ -1493,12 +1522,6 @@ def async_evaluate(
             if not self.enable_monitoring or not self.enable_evaluations:
                 return
```
alanzhang25, line 1523: can u remove background var from async_eval

src/judgeval/local_eval_queue.py
alanzhang25, line 1: what was alternative with multiprocessing

alanzhang25 (changes requested):
small things

src/judgeval/local_eval_queue.py
Mandolaro: we spin up a new process but that is lowkey a bad idea since it might create an orphaned process when their code fails

src/judgeval/__init__.py
Mandolaro: just make it easier to import

alanzhang25 (approved):
lgtm

PR #499, opened 2025-08-04
Add threshold to Prompt Scorer

adivate2021:


alanzhang25 (changes requested):
nice! can we move the prompt scorer tests to its own file and then add a new e2e test for create/get testing that all these things are working

alanzhang25 (approved):
lgtm

PR #505, opened 2025-08-08
Azure + Ollama support (JUD-1713)

yifan1207:
- add azure openai and azure async openi support
- add ollama support (no cost calcualtion support though)


src/judgeval/common/tracer/core.py
```diff
@@ -2020,9 +2053,13 @@ def _format_output_data(
             "OpenAI parsed chat completion not found"
         )
 
-        if isinstance(client, (openai_OpenAI, openai_AsyncOpenAI)):
+        if isinstance(client, (openai_OpenAI, openai_AsyncOpenAI, openai_AzureOpenAI, openai_AsyncAzureOpenAI)):
             if isinstance(response, openai_ChatCompletion):
-                model_name = response.model
+                #added this to diff openai vs azure openai dont know if needed though
+                if isinstance(client, (openai_AzureOpenAI, openai_AsyncAzureOpenAI)):
+                    model_name = "(azure-openai)" + response.model
```
alanzhang25: the model_name is used for calculating cost for litellm? you should check if the pricing differs

src/judgeval/common/tracer/core.py
```diff
@@ -2157,6 +2197,30 @@ def _format_output_data(
...
                 cache_creation_input_tokens,
             )
+        
+    if HAS_OLLAMA:
+        from judgeval.common.tracer.providers import ollama_AsyncClient, ollama_Client
+
+        assert ollama_Client is not None, "Ollama Client not found"
+        assert ollama_AsyncClient is not None, "Ollama Async Client not found"
+        if isinstance(client, (ollama_Client, ollama_AsyncClient)):
+            model_name = "ollama/" + getattr(response, 'model', 'unknown')
```
alanzhang25, line 2205: same here, check if litellm supports olama cost checking and if this is how they format the names for diff models

src/judgeval/common/tracer/core.py
```diff
@@ -2157,6 +2197,30 @@ def _format_output_data(
...
+            prompt_tokens = getattr(response, 'prompt_eval_count', None) or 0
+            completion_tokens = getattr(response, 'eval_count', None) or 0
```
alanzhang25, line 2207: ollama labels their tokens like this?

src/judgeval/common/tracer/providers.py
```diff
@@ -86,6 +88,18 @@
 except ImportError:
     pass
 
+HAS_OLLAMA = False
+ollama_Client = None
+ollama_AsyncClient = None
+
+try:
+    from ollama import Client, AsyncClient, chat
```
alanzhang25: can delete chat right? github lint failing

src/judgeval/common/tracer/core.py
yifan1207: yes

src/judgeval/common/tracer/core.py
yifan1207: I think the issue is azaure does not show its pricing since its mostly enterprise contracts and the compute units based. So I think the best way might be just to use the openai cost as an estimate? or maybe just dont show it

src/judgeval/common/tracer/core.py
yifan1207: I think this is a similar issue I think to calculate ollama costs we'd need the user to return their compute usage but by defualt ollama does not return that. let me remove this. maybe we can default to N/A?

src/judgeval/common/tracer/providers.py
yifan1207: yess. will do

PR #508, opened 2025-08-10
forge fix (JUD-1707)

yifan1207:
-do not trace the spans with names that start with __ and ends with __ the metadata fucntions in langchian


JCamyre (approved):
LGTM!

PR #510, opened 2025-08-11
Generator Wrapper and Tracer Refactor (JUD-1719) 

Mandolaro:
## 📝 Summary

- [ ] 1. Add Generator Wrapper
- [ ] 2. Refactor Tracer Components

## ✅ Checklist

### DIFF
--- a/src/judgeval/common/tracer/core.py
+++ b/src/judgeval/common/tracer/core.py
@@ -1230,129 +1230,207 @@ class Tracer:
         except Exception:
             return func
 
-        if asyncio.iscoroutinefunction(func):
+        def _record_span_data(span, args, kwargs):
+            """Helper function to record inputs, agent info, and state on a span."""
+            # Get class and agent info
+            class_name = None
+            agent_name = None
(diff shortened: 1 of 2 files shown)


src/judgeval/common/tracer/core.py
```diff
@@ -1230,256 +1230,226 @@ def observe(
...
-
-                        # Capture state after execution
-                        self._conditionally_capture_and_record_state(
-                            span, args, is_before=False
-                        )
+                            # Handle automatic trace creation and span execution
+                            item = _execute_with_auto_trace_creation(
+                                original_span_name,
+                                span_type,
+                                lambda: async_generator.__anext__(),
```
abhishekg999, line 1428: will it have the same async behavior when calling it within the sync function _execute_with_auto_trace_creation?

This r seem a bit strangs since the underlying implementation in my mind of traced_async_generator and traced_generator should be identical other than the await, but one the underlying _execute_with_auto_trace_creation is sync.

PR #514, opened 2025-08-15
Otel judgment.agent

justinsheu:
## 📝 Summary

- [ ] 1. ...

## ✅ Checklist

### DIFF
--- a/src/judgeval/tracer/__init__.py
+++ b/src/judgeval/tracer/__init__.py
@@ -17,6 +17,7 @@ from typing import (
     TypeVar,
     overload,
     Literal,
+    TypedDict,
 )
 from functools import partial
 from warnings import warn
@@ -63,9 +64,19 @@ C = TypeVar("C", bound=Callable)
 Cls = TypeVar("Cls", bound=Type)
 ApiClient = TypeVar("ApiClient", bound=Any)
 
-_current_agent_context: ContextVar[
-    Optional[Dict[str, str | bool | Dict | List | None]]
-] = ContextVar("current_agent_context", default=None)
+
+class AgentContext(TypedDict):
+    agent_id: str
+    class_name: str | None
+    track_state: bool
+    track_attributes: List[str] | None
+    field_mappings: Dict[str, str]
+    instance: Any
+
+
+_current_agent_context: ContextVar[Optional[AgentContext]] = ContextVar(
+    "current_agent_context", default=None
+)
 
 _current_cost_context: ContextVar[Optional[Dict[str, float]]] = ContextVar(
     "current_cost_context", default=None
@@ -270,24 +281,10 @@ class Tracer:
         current_agent_context = _current_agent_context.get()
 
         if current_agent_context and current_agent_context.get("track_state"):
-            instance_untyped = current_agent_context.get("instance", None)
-            instance = (
-                instance_untyped if isinstance(instance_untyped, object) else None
-            )
-            track_attributes_untyped = current_agent_context.get(
-                "track_attributes", None
-            )
-            track_attributes: List | None = (
-                track_attributes_untyped
-                if isinstance(track_attributes_untyped, list)
-                else None
-            )
-            field_mappings_untyped = current_agent_context.get("field_mappings", {})
-            field_mappings: Dict[str, str] = (
-                field_mappings_untyped
-                if isinstance(field_mappings_untyped, dict)
-                else {}
-            )
+            instance = current_agent_context.get("instance")
+            track_attributes = current_agent_context.get("track_attributes")
+            field_mappings = current_agent_context.get("field_mappings", {})
+
             if track_attributes is not None:
                 attributes = {
                     field_mappings.get(attr, attr): getattr(instance, attr, None)
@@ -299,17 +296,12 @@ class Tracer:
                     for k, v in instance.__dict__.items()
                     if not k.startswith("_")
                 }
-
-            if record_point == "before":
-                span.set_attribute(
-                    AttributeKeys.JUDGMENT_STATE_BEFORE,
-                    safe_serialize(attributes),
-                )
-            else:
-                span.set_attribute(
-                    AttributeKeys.JUDGMENT_STATE_AFTER,
-                    safe_serialize(attributes),
-                )
+            span.set_attribute(
+                AttributeKeys.JUDGMENT_STATE_BEFORE
+                if record_point == "before"
+                else AttributeKeys.JUDGMENT_STATE_AFTER,
+                safe_serialize(attributes),
+            )
 
     def _wrap_sync(
         self, f: Callable, name: Optional[str], attributes: Optional[Dict[str, Any]]
--- a/src/judgeval/tracer/__init__.py
+++ b/src/judgeval/tracer/__init__.py
@@ -50,6 +50,7 @@ from judgeval.tracer.managers import sync_span_context
 from judgeval.utils.serialize import safe_serialize
 from judgeval.version import get_version
 from judgeval.warnings import JudgmentWarning
+from judgeval.exceptions import JudgmentRuntimeError
 
 from judgeval.tracer.exporters import JudgmentSpanExporter
 from judgeval.tracer.keys import AttributeKeys, ResourceKeys
@@ -490,7 +491,7 @@ class Tracer:
 
             if identifier:
                 if not class_name or not instance or not isinstance(instance, object):
-                    raise Exception(
+                    raise JudgmentRuntimeError(
                         "'identifier' is set but no class name or instance is available. 'identifier' can only be specified when using the agent() decorator on a class method."
                     )
                 if (
@@ -501,7 +502,7 @@ class Tracer:
                     instance_name = str(getattr(instance, identifier))
                     agent_context["instance_name"] = instance_name
                 else:
-                    raise Exception(
+                    raise JudgmentRuntimeError(
                         f"Attribute {identifier} does not exist for {class_name}. Check your agent() decorator."
                     )
 
--- a/src/judgeval/tracer/__init__.py
+++ b/src/judgeval/tracer/__init__.py
@@ -3,7 +3,6 @@ from contextvars import ContextVar
 import functools
 import inspect
 import random
-import uuid
 from typing import (
     Any,
     Union,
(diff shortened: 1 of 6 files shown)


src/judgeval/tracer/__init__.py
```diff
@@ -202,13 +213,112 @@ def get_current_span(self):
...
+        return _current_agent_context
+
+    def get_current_cost_context(self):
+        return _current_cost_context
+
+    def add_cost_to_current_context(self, cost: float) -> None:
+        """Add cost to the current cost context and update span attribute."""
+        current_cost_context = _current_cost_context.get()
+        if current_cost_context is not None:
+            current_cumulative_cost = current_cost_context.get("cumulative_cost", 0.0)
```
abhishekg999, line 239: we never set the cumulative_cost?

src/judgeval/tracer/__init__.py
```diff
@@ -202,13 +213,112 @@ def get_current_span(self):
...
+            new_cumulative_cost = float(current_cumulative_cost) + cost
+            current_cost_context["cumulative_cost"] = new_cumulative_cost
```
abhishekg999, line 241: will this work this is the context var object here right?

src/judgeval/tracer/managers.py
```diff
@@ -30,8 +44,20 @@ async def async_span_context(
...
+    cost_context = {"cumulative_cost": 0.0}
+
+    cost_token = current_cost_context.set(cost_context)
+
+    try:
+        with tracer.get_tracer().start_as_current_span(
+            name=name,
+            attributes=span_attributes,
+        ) as span:
+            span.set_attribute(AttributeKeys.JUDGMENT_CUMULATIVE_LLM_COST, 0.0)
```
abhishekg999, line 60: why do we store cumulative_cost on every single span? We will have spans that inevitably have no associattion with llm cost correct. Im just curious the decision here

src/judgeval/tracer/managers.py
```diff
@@ -16,11 +17,24 @@ def sync_span_context(
...
+    cost_context = {"cumulative_cost": 0.0}
+
+    cost_token = current_cost_context.set(cost_context)
+
+    try:
+        with tracer.get_tracer().start_as_current_span(
+            name=name,
+            attributes=span_attributes,
+        ) as span:
+            # Set initial cumulative cost attribute
```
abhishekg999, line 33: (similar) why do we store cumulative_cost on every single span? We will have spans that inevitably have no associattion with llm cost correct. Im just curious the decision here

src/judgeval/tracer/keys.py
```diff
@@ -27,6 +37,10 @@ class AttributeKeys:
     GEN_AI_REQUEST_MAX_TOKENS = gen_ai_attributes.GEN_AI_REQUEST_MAX_TOKENS
     GEN_AI_RESPONSE_FINISH_REASONS = gen_ai_attributes.GEN_AI_RESPONSE_FINISH_REASONS
 
+    # GenAI-specific attributes (custom namespace)
+    GEN_AI_USAGE_TOTAL_TOKENS = "gen_ai.usage.total_tokens"
```
abhishekg999: we already have GEN_AI_USAGE_INPUT_TOKENS and GEN_AI_USAGE_OUTPUT_TOKENS as proper keys, what is this total tokens

src/judgeval/tracer/__init__.py
```diff
@@ -282,6 +396,149 @@ def observe(
...
+            agent_context["track_state"] = track_state
+            agent_context["track_attributes"] = track_attributes
+            agent_context["field_mappings"] = field_mappings
+
+            instance = args[0] if args else None
+            agent_context["instance"] = instance
+
+            if identifier:
+                if not class_name or not instance or not isinstance(instance, object):
+                    raise Exception(
```
abhishekg999: maybe raise JudgmentRuntimeError

src/judgeval/tracer/__init__.py
```diff
@@ -282,6 +396,149 @@ def observe(
...
+
+            current_agent_context = _current_agent_context.get()
+            if current_agent_context and "agent_id" in current_agent_context:
+                agent_context["parent_agent_id"] = current_agent_context["agent_id"]
+
+            agent_context["is_agent_entry_point"] = True
+            token = _current_agent_context.set(agent_context)
+            return token
+
+        def _wrap_with_agent_context(f: Callable):
```
abhishekg999: hm i would say to move these stuff to managers.py since managers.py was originally created to manage this sort of span and context handling. 

I would suggest still doing that and handle the token creation and reset that way by passing the token or the context if needed to the manager as an argument.

abhishekg999: what do you think here? Is there a reason we added this here and not along with the other span stuff in managers.py?

src/judgeval/tracer/__init__.py
```diff
@@ -202,13 +213,112 @@ def get_current_span(self):
...
+                )
+                current_agent_context["is_agent_entry_point"] = (
+                    False  # only true for entry point to agent
+                )
+
+    def record_instance_state(self, record_point: Literal["before", "after"], span):
+        current_agent_context = _current_agent_context.get()
+
+        if current_agent_context and current_agent_context.get("track_state"):
+            instance_untyped = current_agent_context.get("instance", None)
```
abhishekg999: (nitpick) this is like 50 lines of code for literally just accessing 4 attributes of the object can this be better? 
If we decide to continue using _current_agent_context, can we typedef it as a TypedDict and define what it should have.

src/judgeval/tracer/__init__.py
```diff
@@ -202,13 +213,112 @@ def get_current_span(self):
...
+            span = get_current_span()
+            if span and span.is_recording():
+                span.set_attribute(
+                    AttributeKeys.JUDGMENT_CUMULATIVE_LLM_COST, new_cumulative_cost
+                )
+
+    def add_agent_attributes_to_span(self, span):
+        """Add agent ID, class name, and instance name to span if they exist in context"""
+        current_agent_context = _current_agent_context.get()
+        if current_agent_context:
```
abhishekg999, line 252: (very small nitpick) for readability can we early return or something if the current_agent_context doesnt exist?

something like 

```py
current_agent_context = _current_agent_context.get()
if not current_agent_context: return
# continue with remaining set attributes
...
```

src/judgeval/tracer/__init__.py
```diff
@@ -59,6 +62,14 @@
 Cls = TypeVar("Cls", bound=Type)
 ApiClient = TypeVar("ApiClient", bound=Any)
 
+_current_agent_context: ContextVar[
```
abhishekg999: Make these local to tracer if needed, but also why context vars needed for this? Can we accumulate this information in our db? 

LMK what the idea here was with these two context vars ideally less contexts to manage the better

src/judgeval/tracer/__init__.py
justinsheu: in managers.py cumulative_cost should be set and initialized to 0

src/judgeval/tracer/__init__.py
justinsheu: this is the context var dictionary referenced by the context var (_current_cost_context.get()), so when we modify it, since it's by reference, this works (at least from the testing i've done)

justinsheu: maybe var name could be made more clear?

src/judgeval/tracer/keys.py
justinsheu: good point, we can just sum up input and output token to get total, i just added this cause the openai response also returns total_tokens but i think this is unnecessary

src/judgeval/tracer/managers.py
justinsheu: i think the idea behind this was that since we're recursively rolling costs upwards through span contexts a lot of spans have costs that are the sum of their children's cumulative costs. i was thinking in the future if we want to give users a way to know how much a specific span incurred cost-wise through itself and its children this would be an easy way to do that

justinsheu: in memory we're already tracking the costs of each child span so i thought might as well track it, is there a substantial performance for this operation (maybe on the db trigger for populating the cumulative_llm_cost column)?

src/judgeval/tracer/__init__.py
adivate2021: The reason for the context vars is that the children spans need the information about the agent, which comes from the parent span (since many spans will refer to the same agent id). I'm not sure if there's a more convenient way to do this

src/judgeval/tracer/__init__.py
adivate2021: just moved to managers.py

src/judgeval/tracer/keys.py
justinsheu: resolved

src/judgeval/tracer/__init__.py
justinsheu: condensed

src/judgeval/tracer/managers.py
justinsheu: see comment below

src/judgeval/tracer/__init__.py
```diff
@@ -60,6 +66,24 @@
...
+class AgentContext(TypedDict):
+    agent_id: str
+    class_name: str | None
+    track_state: bool
+    track_attributes: List[str] | None
+    field_mappings: Dict[str, str]
+    instance: Any
+
+
+_current_agent_context: ContextVar[Optional[AgentContext]] = ContextVar(
```
abhishekg999: oops I think I forgot to send my comment for this but, Can we not maintain global context vars here. We are trying to isolate Tracer instances.

1. Can these be instance variables associated to the tracer
2. Ideally we can already aggregate this in our database, why do we need this aggregation on the client at all? I feel the responsiblity here of the tracer should be as clear and simple as possible and we shoild not be introduction new context vars we need to manually manage for each new thing we want to add.

adivate2021: Just made them instance variables, as we discussed we can change the logic to just store raw data once we move to Clickhouse but for now should work as temporary solution

abhishekg999 (approved):

PR #520, opened 2025-08-21
dashboard otel

justinsheu:
## 📝 Summary

- [ ] 1. ...

## ✅ Checklist

### DIFF
--- a/src/judgeval/tracer/__init__.py
+++ b/src/judgeval/tracer/__init__.py
@@ -323,7 +323,6 @@ class Tracer:
             with sync_span_context(self, n, attributes) as span:
                 self.add_agent_attributes_to_span(span)
                 self.record_instance_state("before", span)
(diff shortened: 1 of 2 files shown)


src/judgeval/tracer/__init__.py
```diff
@@ -320,6 +323,7 @@ def wrapper(*args, **kwargs):
             with sync_span_context(self, n, attributes) as span:
                 self.add_agent_attributes_to_span(span)
                 self.record_instance_state("before", span)
+                span.set_attribute(AttributeKeys.JUDGMENT_OFFLINE_MODE, False)
```
abhishekg999: 1 question: so once again is this being set on every single request? 
I think for these special attributes, we should try to have it as opt-in as possible. meaning we should have a reasonable default where if offline mode is not specified then by default we know its online.

This is similar to like the other principles we are trying to keep in the otel tracer to have it behave as similar to other trace providers as possible.

PR #525, opened 2025-08-29
Add llm streaming (JUD-1720)

Mandolaro:
## 📝 Summary

- [ ] 1. Add streaming support for OTel

## ✅ Checklist

### DIFF
--- a/src/judgeval/tracer/llm/__init__.py
+++ b/src/judgeval/tracer/llm/__init__.py
@@ -42,6 +42,425 @@ def cost_per_token(
         return None, None
 
 
+class TracedGenerator:
+    """Generator wrapper that adds OpenTelemetry tracing without consuming the stream."""
+
+    def __init__(
+        self, tracer: "Tracer", generator, client: ApiClient, span, model_name: str = ""
+    ):
+        self.tracer = tracer
+        self.generator = generator
+        self.client = client
+        self.span = span
+        self.accumulated_content = ""
+        self.model_name = model_name
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        try:
+            chunk = next(self.generator)
+
+            content = self._extract_content(chunk)
+            if content:
+                self.accumulated_content += content
+            self._process_chunk_usage(chunk)
+
+            return chunk
+
+        except StopIteration:
+            self._finalize_span()
+            raise
+        except Exception as e:
+            self.span.record_exception(e)
+            self.span.end()
+            raise
+
+    def _extract_content(self, chunk) -> str:
+        """Extract content from streaming chunk based on provider."""
+        if HAS_OPENAI:
+            from judgeval.tracer.llm.providers import openai_OpenAI, openai_AsyncOpenAI
+
+            if isinstance(self.client, (openai_OpenAI, openai_AsyncOpenAI)):
+                if (
+                    hasattr(chunk, "choices")
+                    and chunk.choices
+                    and hasattr(chunk.choices[0], "delta")
+                ):
+                    delta_content = getattr(chunk.choices[0].delta, "content", None)
+                    if delta_content:
+                        return delta_content
+
+        if HAS_ANTHROPIC:
+            from judgeval.tracer.llm.providers import (
+                anthropic_Anthropic,
+                anthropic_AsyncAnthropic,
+            )
+
+            if isinstance(self.client, (anthropic_Anthropic, anthropic_AsyncAnthropic)):
+                if hasattr(chunk, "type") and chunk.type == "content_block_delta":
+                    if hasattr(chunk, "delta") and hasattr(chunk.delta, "text"):
+                        return chunk.delta.text or ""
+                elif hasattr(chunk, "delta") and hasattr(chunk.delta, "text"):
+                    return chunk.delta.text or ""
+                elif hasattr(chunk, "text"):
+                    return chunk.text or ""
+
+        if HAS_TOGETHER:
+            from judgeval.tracer.llm.providers import (
+                together_Together,
+                together_AsyncTogether,
+            )
+
+            if isinstance(self.client, (together_Together, together_AsyncTogether)):
+                if hasattr(chunk, "choices") and chunk.choices:
+                    choice = chunk.choices[0]
+                    if hasattr(choice, "delta") and hasattr(choice.delta, "content"):
+                        return choice.delta.content or ""
+
+        if HAS_GROQ:
+            from judgeval.tracer.llm.providers import groq_Groq, groq_AsyncGroq
+
+            if isinstance(self.client, (groq_Groq, groq_AsyncGroq)):
+                if hasattr(chunk, "choices") and chunk.choices:
+                    choice = chunk.choices[0]
+                    if hasattr(choice, "delta") and hasattr(choice.delta, "content"):
+                        return choice.delta.content or ""
+
+        return ""
+
+    def _process_chunk_usage(self, chunk):
+        """Process usage data from streaming chunks based on provider."""
+        usage_data = None
+
+        if HAS_ANTHROPIC:
+            from judgeval.tracer.llm.providers import (
+                anthropic_Anthropic,
+                anthropic_AsyncAnthropic,
+            )
+
+            if isinstance(self.client, (anthropic_Anthropic, anthropic_AsyncAnthropic)):
+                if hasattr(chunk, "type"):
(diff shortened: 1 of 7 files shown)


alanzhang25 (commented):
/claude review

src/judgeval/tracer/llm/__init__.py
```diff
@@ -42,6 +42,322 @@ def cost_per_token(
...
+        prompt_tokens = getattr(usage_data, "prompt_tokens", 0) or 0
+    if not completion_tokens and hasattr(usage_data, "completion_tokens"):
+        completion_tokens = getattr(usage_data, "completion_tokens", 0) or 0
+
+    if hasattr(usage_data, "cache_read_input_tokens"):
+        cache_read_input_tokens = getattr(usage_data, "cache_read_input_tokens", 0) or 0
+    if hasattr(usage_data, "cache_creation_input_tokens"):
+        cache_creation_input_tokens = (
+            getattr(usage_data, "cache_creation_input_tokens", 0) or 0
+        )
```
alanzhang25: doesnt this assume that every clients usage_data has the same naming for tokens? i dont believe this is true

src/judgeval/tracer/llm/__init__.py
```diff
@@ -42,6 +42,322 @@ def cost_per_token(
...
+            prompt_tokens,
+            completion_tokens,
+            cache_read_input_tokens,
+            cache_creation_input_tokens,
+        )
+        _set_usage_attributes(span, usage, tracer)
+
+
+def _set_usage_attributes(span, usage: TraceUsage, tracer: "Tracer"):
+    """Set usage attributes on the span for non-streaming responses."""
```
alanzhang25, line 861: this is misleading because this is eventually used by stremaing responses?

I feel that passing in Tracer obj at this level isnt right either

alanzhang25 (changes requested):
Can you also add e2e for streaming llm cases

src/judgeval/tracer/llm/__init__.py
Mandolaro: are you saying this for the cache?

src/judgeval/tracer/llm/__init__.py
Mandolaro: no we need this for the span also

src/judgeval/tracer/llm/__init__.py
```diff
@@ -42,6 +42,322 @@ def cost_per_token(
         return None, None
 
 
+class _TracedGeneratorBase:
```
abhishekg999, line 546: for consistancy can we use slots for these types of objects + proper type declarations refer to Tracer

src/judgeval/tracer/llm/__init__.py
```diff
@@ -42,6 +42,322 @@ def cost_per_token(
...
+
+    def __del__(self):
+        """
+        Fallback cleanup for unclosed spans. Note: __del__ is not guaranteed to be called
+        in all situations (e.g., reference cycles, program exit), so this should not be
+        relied upon as the primary cleanup mechanism.
+        """
+        if hasattr(self, "span") and self.span:
+            try:
+                self._finalize_span()
```
abhishekg999, line 614: why does deleting the object finalize the span? those seem like conceptually different things, we already specify the finalize span in the StopIteration which I think makes sense to me

src/judgeval/tracer/llm/__init__.py
```diff
@@ -42,6 +42,322 @@ def cost_per_token(
...
+        """Finalize the span by setting completion content and ending it."""
+        if hasattr(self, "span") and self.span:
+            set_span_attribute(
+                self.span, AttributeKeys.GEN_AI_COMPLETION, self.accumulated_content
+            )
+            self.span.end()
+            self.span = None
+
+
+class TracedGenerator(_TracedGeneratorBase):
```
abhishekg999, line 630: same comment on slots please

src/judgeval/tracer/llm/__init__.py
```diff
@@ -42,6 +42,322 @@ def cost_per_token(
...
+    """Generator wrapper that adds OpenTelemetry tracing without consuming the stream."""
+
+    def __init__(
+        self, tracer: "Tracer", generator, client: ApiClient, span, model_name: str = ""
```
abhishekg999: (nitpick)
can we `from __future__ import annotations`
on the top of thsis file and remove the quotes on Tracer.

abhishekg999: Why is model_name default none? Is that a fine default?

src/judgeval/tracer/llm/__init__.py
```diff
@@ -42,6 +42,322 @@ def cost_per_token(
...
+        except StopIteration:
+            self._finalize_span()
+            raise
+        except Exception as e:
+            self.span.record_exception(e)
+            self.span.end()
+            raise
+
+
+class TracedAsyncGenerator(_TracedGeneratorBase):
```
abhishekg999, line 672: similar on slots

src/judgeval/tracer/llm/__init__.py
```diff
@@ -42,6 +42,322 @@ def cost_per_token(
...
+            self.span.end()
+            raise
+
+
+class TracedSyncContextManager:
+    """Sync context manager wrapper for streaming methods."""
+
+    def __init__(
+        self,
+        tracer: "Tracer",
```
abhishekg999: (nit) similar on slots and future import

src/judgeval/tracer/llm/__init__.py
```diff
@@ -42,6 +42,322 @@ def cost_per_token(
...
+        self.accumulated_content = ""
+        self.model_name = model_name
+
+    def _extract_content(self, chunk) -> str:
+        """Extract content from streaming chunk based on provider."""
+        if HAS_OPENAI:
+            from judgeval.tracer.llm.providers import openai_OpenAI, openai_AsyncOpenAI
+
+            if isinstance(self.client, (openai_OpenAI, openai_AsyncOpenAI)):
+                if (
```
abhishekg999: Thius is alot of duplicated code to the existing implementation, IK there is a different handling of delta but can this be cleaned up possibly?

src/judgeval/tracer/llm/__init__.py
```diff
@@ -42,6 +42,322 @@ def cost_per_token(
...
+        self.stream = await self.context_manager.__aenter__()
+        return TracedAsyncGenerator(
+            self.tracer, self.stream, self.client, self.span, self.model_name
+        )
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb):
+        return await self.context_manager.__aexit__(exc_type, exc_val, exc_tb)
+
+    def __del__(self):
+        if hasattr(self, "span") and self.span:
```
abhishekg999: Why are we hasattr? Can I get an explanation of this, In what case wouldnt self have span as an attribute.

src/judgeval/tracer/llm/__init__.py
Mandolaro: this is just when user dont fully consume the generator, if user dont fully close the generator we still need to finalize it

src/judgeval/tracer/llm/__init__.py
Mandolaro: like this just follow the current format output code, it depends if you want to make it more simplify.
But the delta stuff change has to be custom from the normal code

src/judgeval/tracer/llm/__init__.py
```diff
@@ -42,6 +42,501 @@ def cost_per_token(
...
+        "tracer",
+        "client",
+        "span",
+        "accumulated_content",
+        "model_name",
+    )
+
+    tracer: Tracer
+    client: ApiClient
+    span: Any
```
abhishekg999: Why is span Any, This will be an otel span right?

src/judgeval/tracer/llm/__init__.py
```diff
@@ -42,6 +42,501 @@ def cost_per_token(
...
+            self.span.end()
+            raise
+
+
+class TracedAsyncGenerator(_TracedGeneratorBase):
+    """Async generator wrapper that adds OpenTelemetry tracing without consuming the stream."""
+
+    __slots__ = ("async_generator",)
+
+    async_generator: Any
```
abhishekg999: Why is everything typed any

PR #528, opened 2025-09-01
Add back generator code

Mandolaro:
## 📝 Summary

- [ ] 1. Add generator logic

## ✅ Checklist

### DIFF
--- a/src/judgeval/tracer/__init__.py
+++ b/src/judgeval/tracer/__init__.py
@@ -51,6 +51,7 @@ from judgeval.scorers.base_scorer import BaseScorer
 from judgeval.tracer.constants import JUDGEVAL_TRACER_INSTRUMENTING_MODULE_NAME
 from judgeval.tracer.managers import (
     sync_span_context,
+    async_span_context,
     sync_agent_context,
     async_agent_context,
 )
@@ -572,7 +573,7 @@ class Tracer:
         @functools.wraps(f)
         async def wrapper(*args, **kwargs):
             n = name or f.__qualname__
-            with sync_span_context(self, n, attributes) as span:
+            async with async_span_context(self, n, attributes) as span:
                 self.add_agent_attributes_to_span(span)
                 self.record_instance_state("before", span)
                 try:
@@ -605,7 +606,7 @@ class Tracer:
                                     **(attributes or {}),
                                 }
 
-                                with sync_span_context(
+                                async with async_span_context(
                                     self, yield_span_name, yield_attributes
                                 ) as yield_span:
                                     self.add_agent_attributes_to_span(yield_span)
@@ -654,7 +655,7 @@ class Tracer:
         async def wrapper(*args, **kwargs):
             n = name or f.__qualname__
 
-            with sync_span_context(self, n, attributes) as main_span:
+            async with async_span_context(self, n, attributes) as main_span:
                 self.add_agent_attributes_to_span(main_span)
                 self.record_instance_state("before", main_span)
 
@@ -682,7 +683,7 @@ class Tracer:
                                     **(attributes or {}),
                                 }
 
-                                with sync_span_context(
+                                async with async_span_context(
                                     self, yield_span_name, yield_attributes
                                 ) as yield_span:
                                     self.add_agent_attributes_to_span(yield_span)
--- a/src/judgeval/tracer/__init__.py
+++ b/src/judgeval/tracer/__init__.py
@@ -419,9 +419,9 @@ class Tracer:
                     except StopIteration:
                         # Mark span as cancelled so it won't be exported
                         self.judgment_processor.set_internal_attribute(
-                            yield_span.get_span_context(),
-                            InternalAttributeKeys.CANCELLED,
-                            True,
+                            span_context=yield_span.get_span_context(),
+                            key=InternalAttributeKeys.CANCELLED,
+                            value=True,
                         )
                         break
 
@@ -463,9 +463,9 @@ class Tracer:
                     except StopAsyncIteration:
                         # Mark span as cancelled so it won't be exported
                         self.judgment_processor.set_internal_attribute(
-                            yield_span.get_span_context(),
-                            InternalAttributeKeys.CANCELLED,
-                            True,
+                            span_context=yield_span.get_span_context(),
+                            key=InternalAttributeKeys.CANCELLED,
+                            value=True,
                         )
                         break
 
--- a/src/judgeval/tracer/__init__.py
+++ b/src/judgeval/tracer/__init__.py
@@ -793,7 +793,7 @@ class Tracer:
             **(attributes or {}),
         }
 
-        if inspect.iscoroutinefunction(func):
+        if inspect.iscoroutinefunction(func) or inspect.isasyncgenfunction(func):
             return self._wrap_async(func, name, func_attributes, scorer_config)
         else:
             return self._wrap_sync(func, name, func_attributes, scorer_config)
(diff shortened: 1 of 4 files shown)


src/judgeval/tracer/processors/__init__.py
```diff
@@ -58,6 +58,11 @@ def emit_partial(self) -> None:
         if not isinstance(current_span, ReadableSpan):
             return
 
+        # Check if partial emit is disabled for this span (e.g., generator yields)
+        attributes = current_span.attributes or {}
+        if attributes.get("judgment.span.disable_partial_emit"):
```
abhishekg999: we shouldnt store this runtime library used state in span attributes since it will be emitted to all exporters and its just noise.

src/judgeval/tracer/processors/__init__.py
```diff
@@ -84,6 +89,15 @@ def emit_partial(self) -> None:
         super().on_end(partial_span)
 
     def on_end(self, span: ReadableSpan) -> None:
+        # Check if span is marked as cancelled (due to StopIteration)
```
abhishekg999: I dont understand this comment, stop iteration is a normal exception to be raised its how the iteration ends. why would that mean cancelled.

src/judgeval/tracer/processors/__init__.py
```diff
@@ -84,6 +89,15 @@ def emit_partial(self) -> None:
...
+        attributes = span.attributes or {}
+        if attributes.get("judgment.span.cancelled"):
```
abhishekg999: similarly we shouldnt store this data as an attribute i dont think

I think there is a small thing we should think about where some of this data lives on the exporter that adds it (But this is a different thing)

src/judgeval/tracer/__init__.py
```diff
@@ -421,11 +431,128 @@
...
+                    self.record_instance_state("after", main_span)
+
+                    return traced_generator()
+
+                except Exception as user_exc:
+                    main_span.record_exception(user_exc)
+                    main_span.set_status(Status(StatusCode.ERROR, str(user_exc)))
+                    raise
 
         return wrapper
```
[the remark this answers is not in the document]
abhishekg999: @Mandolaro can you make the cumulative diff of this pr like atleast -300 lol

src/judgeval/tracer/__init__.py
```diff
@@ -459,11 +592,129 @@
...
+                    self.record_instance_state("after", main_span)
+
+                    return traced_async_generator()
+
+                except Exception as user_exc:
+                    main_span.record_exception(user_exc)
+                    main_span.set_status(Status(StatusCode.ERROR, str(user_exc)))
+                    raise
 
         return wrapper
```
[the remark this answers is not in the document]
abhishekg999: @Mandolaro

src/judgeval/tracer/__init__.py
```diff
@@ -390,13 +395,99 @@ def _set_pending_trace_eval(
...
+
+                async with async_span_context(
+                    self, yield_span_name, yield_attributes, disable_partial_emit=True
+                ) as yield_span:
+                    self.add_agent_attributes_to_span(yield_span)
+
+                    try:
+                        value = await async_generator.__anext__()
+                    except StopAsyncIteration:
+                        # Mark span as cancelled so it won't be exported
```
abhishekg999, line 464: did you test this? stop iteration is a normal exception to be raised im pretty sure why do we want to cancel these spans

src/judgeval/tracer/__init__.py
```diff
@@ -487,31 +699,104 @@ def observe(
         scorer_config: TraceScorerConfig | None = None,
     ) -> Callable[[C], C]: ...
 
+    @overload
     def observe(
         self,
-        func: Callable | None = None,
+        func: Generator[Any, None, None],
```
abhishekg999: this messes up the customers implementation typing right? look at the above observe overloads, we dont want to override anything that users define. Can the type arguments of the generator be type variables

src/judgeval/tracer/__init__.py
```diff
@@ -487,31 +699,104 @@ def observe(
...
+                attributes=attributes,
+                scorer_config=scorer_config,
             )
 
         if not self.enable_monitoring:
             return func
 
-        name = func.__qualname__
-        attributes: Dict[str, Any] = {
+        # Handle direct iterator/generator objects that are already instantiated
```
abhishekg999: Why would we handle an object thats already instantiated. I dont think that can even be decorated in the tradional sense. Can we just follow the pattern whre Callables are the ones that are being decorated

src/judgeval/tracer/__init__.py
```diff
@@ -487,31 +699,104 @@ def observe(
...
+        if inspect.isgenerator(func) or isinstance(func, Iterator):
```
abhishekg999: (nit) can we follow the pattern of breaking up the underlying logic of handling a speicifc type of object within its own function. similar to _wrap_sync and _wrap_async

src/judgeval/tracer/managers.py
```diff
@@ -32,6 +33,8 @@ def sync_span_context(
             attributes=span_attributes,
         ) as span:
             set_span_attribute(span, AttributeKeys.JUDGMENT_CUMULATIVE_LLM_COST, 0.0)
+            if disable_partial_emit:
+                span.set_attribute("judgment.span.disable_partial_emit", True)
```
abhishekg999: similar, lets not make this a span attribute, but we can think of another way to maintain this global state. I think we should expose some utility functions within the tracer object that gives a span level store to store arbitrary data to associate to a span might be useful here associating a trace_id to a Dict

abhishekg999 (changes requested):

src/judgeval/tracer/processors/__init__.py
Mandolaro: basically the way it is generator wrapper is coded is that the last iteration of the yield (StopIteration) will spawn an empty span (and this basically mitigate this)

src/judgeval/tracer/__init__.py
Mandolaro: just that these spans of StopIteration is empty and look weird from a UX perspective

src/judgeval/tracer/processors/__init__.py
Mandolaro: I feel like if you reduces amount of state a tracer is tracked, storing in the span is better, but dependent our philosophy of implementing the tracer

src/judgeval/tracer/__init__.py
Mandolaro: resolved

src/judgeval/tracer/processors/__init__.py
```diff
@@ -48,7 +48,39 @@ def __init__(
             max_queue_size=max_queue_size,
             export_timeout_millis=export_timeout_millis,
         )
-        self._span_update_ids: dict[tuple[int, int], int] = {}
+        self._internal_attributes: dict[tuple[int, int], dict[str, Any]] = {}
```
abhishekg999: defaultdict

src/judgeval/tracer/processors/__init__.py
```diff
@@ -106,7 +151,7 @@ def on_end(self, span: ReadableSpan) -> None:
                 instrumentation_scope=span.instrumentation_scope,
             )
 
-            self._span_update_ids.pop(span_key, None)
+            self._cleanup_span_state(span_key)
```
abhishekg999, line 155: nice

abhishekg999 (approved):
lgtm can we just make that small defaultdict change

PR #568, opened 2025-09-22
Aaryan/add model trace prompt scorer

adivate2021:
- add model to prompt scorer
- remove options from prompt scorer

### DIFF
--- a/src/judgeval/scorers/judgeval_scorers/api_scorers/prompt_scorer.py
+++ b/src/judgeval/scorers/judgeval_scorers/api_scorers/prompt_scorer.py
@@ -101,13 +101,6 @@ class BasePromptScorer(ABC, APIScorerConfig):
     judgment_api_key: str = os.getenv("JUDGMENT_API_KEY") or ""
     organization_id: str = os.getenv("JUDGMENT_ORG_ID") or ""
 
-    def __init__(self, *args, **kwargs):
-        raise JudgmentAPIError(
-            status_code=400,
-            detail="Please use the create() class method to instantiate a new PromptScorer or the get() class method to retrieve an existing PromptScorer",
-            response=None,
-        )
-
     @classmethod
     def get(
         cls,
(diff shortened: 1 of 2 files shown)


src/judgeval/api/api_types.py
```diff
@@ -141,7 +141,7 @@ class PromptScorer(TypedDict):
     name: str
     prompt: str
     threshold: float
-    options: NotRequired[Optional[Dict[str, float]]]
+    model: NotRequired[str]
     created_at: NotRequired[Optional[str]]
```
justinsheu, line 145: should Optional be removed from these fields too

src/judgeval/scorers/judgeval_scorers/api_scorers/prompt_scorer.py
```diff
@@ -96,17 +95,8 @@ def scorer_exists(
...
-    2. options for scores on the example
-
-    and uses a judge to execute the evaluation from the system role and classify into one of the options
-    """
-
     score_type: APIScorerType
     prompt: str
-    options: Optional[Dict[str, float]] = None
     judgment_api_key: str = os.getenv("JUDGMENT_API_KEY") or ""
     organization_id: str = os.getenv("JUDGMENT_ORG_ID") or ""
```
justinsheu, line 103: non-blocking: base class should probably prevent direct instantiation

src/judgeval/api/api_types.py
adivate2021: autogenerated, seems like its because model has a default value (but is not optional), whereas the other ones have the optional typing in the backend (with default none). But it should work fine

src/judgeval/scorers/judgeval_scorers/api_scorers/prompt_scorer.py
adivate2021: done

adivate2021: had to revert this change because i dont know how to prevent direct instantiation without preventing instantiation in general, we can make a separate PR to try to do this for all things that use this interface

PR #569, opened 2025-09-23
Tracer initialize API refactor - minor bump

abhishekg999:


src/judgeval/constants.py
```diff
@@ -24,19 +24,15 @@ class APIScorerType(str, Enum):
...
         raise ValueError(f"Invalid scorer type: {value}")
 
 
-UNBOUNDED_SCORERS: Set[APIScorerType] = (
-    set()
-)  # scorers whose scores are not bounded between 0-1
-
-
+# scorers whose scores are not bounded between 0-1
+UNBOUNDED_SCORERS: Set[APIScorerType] = set()
```
adivate2021: not really related to your PR, but this seems to be a pointless constant since all of our scorers are bounded? Could we just remove this, and also remove from validate_threshold in APIScorerConfig?

src/judgeval/tracer/managers.py
```diff
@@ -21,29 +21,17 @@ def sync_span_context(
     if span_attributes is None:
         span_attributes = {}
 
-    current_cost_context = tracer.get_current_cost_context()
```
justinsheu, line 24: are we no longer going to track cumulative cost?

src/judgeval/integrations/langgraph/__init__.py
```diff
@@ -1,789 +1,13 @@
...
-
-        return None
-
-
-__all__ = ["JudgevalCallbackHandler"]
+class Langgraph(ABC):
+    @staticmethod
+    def initialize(otel_only: bool = True):
+        os.environ["LANGSMITH_OTEL_ENABLED"] = "true"
+        os.environ["LANGSMITH_TRACING"] = "true"
```
justinsheu, line 11: i remember back when we were servicing forge having LANGSMITH_TRACING set would cause our tracing to break because of some conflict with langsmith. Is this fixed by not using the callbacks?

src/judgeval/tracer/managers.py
```diff
@@ -21,29 +21,17 @@ def sync_span_context(
     if span_attributes is None:
```
adivate2021, line 20: some valid lint issues in this file

src/judgeval/integrations/langgraph/__init__.py
justinsheu: might be misunderstanding but doesn't this just initialize langsmith tracing? If they want the traces to also show up on our platform don't they now have to manually instrument @observe and wrap() (compared to before where they just had to modify their langgraph initialization and the rest is auto traced). Or does this handle autoinstrumenting judgment tracing with langgraph code?

src/judgeval/tracer/__init__.py
```diff
@@ -134,83 +131,144 @@ def __init__(
...
-        != "false",
-        enable_evaluation: bool = os.getenv(
-            "JUDGMENT_ENABLE_EVALUATIONS", "true"
-        ).lower()
-        != "false",
-        processors: List[SpanProcessor] = [],
+        enable_monitoring: bool = JUDGMENT_ENABLE_MONITORING.lower() == "true",
+        enable_evaluation: bool = JUDGMENT_ENABLE_EVALUATIONS.lower() == "true",
         resource_attributes: Optional[Dict[str, Any]] = None,
+        initialize: bool = True,
```
justinsheu, line 132: in what situation would we not want to initialize the tracer upon instantiation?

src/judgeval/integrations/langgraph/__init__.py
justinsheu: resolved

src/judgeval/tracer/__init__.py
justinsheu: resolved

src/judgeval/tracer/__init__.py
```diff
@@ -86,46 +90,38 @@ class AgentContext(TypedDict):
...
-        # Otel
         "judgment_processor",
-        "processors",
-        "provider",
         "tracer",
-        # Agent
         "agent_context",
-        "cost_context",
+        "_initialized",
+        "_project_name",
```
adivate2021: can remove these variables

src/judgeval/tracer/managers.py
abhishekg999: updated imports

src/judgeval/tracer/managers.py
abhishekg999: correct, will implement cost tracking correctly in the overall cost tracking pr

abhishekg999: will be server side not done in client

src/judgeval/constants.py
abhishekg999: updated good call

pyproject.toml
```diff
@@ -45,12 +44,6 @@ packages = ["src/judgeval"]
 include = ["/src/judgeval", "/src/judgeval/**/*.py"]
 
 [project.optional-dependencies]
-langchain = [
```
abhishekg999, line 48: removed since langchain is controllable purely through env vars

src/judgeval/tracer/__init__.py
abhishekg999: removed

adivate2021 (approved):
lgtm, just fix lint and add a skip for the cost e2es

src/judgeval/scorers/score.py
```diff
@@ -21,7 +21,7 @@
 
 
 async def safe_a_score_example(
-    scorer: Union[ExampleScorer],
+    scorer: [ExampleScorer],
```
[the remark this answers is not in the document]
justinsheu: bump if valid

justinsheu (changes requested):

PR #610, opened 2025-10-20
JUD-2227: Customer ID Passing

alanzhang25:
## 📝 Summary

- [ ] 1. ...

## ✅ Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -51,7 +51,7 @@ jobs:
           cd src
           export JUDGMENT_API_KEY="$JUDGEVAL_GH_JUDGMENT_API_KEY"
           export JUDGMENT_ORG_ID="$JUDGEVAL_GH_JUDGMENT_ORG_ID"
-          uv run pytest tests -n auto
+          uv run pytest tests/tracer/llm -n auto
 
   run-e2e-tests:
     needs: [validate-branch]


src/judgeval/tracer/__init__.py
```diff
@@ -247,14 +250,23 @@ def get_tracer(self):
...
     def set_customer_id(self, customer_id: str) -> None:
         span = self.get_current_span()
+
+        if self.get_current_customer_id():
+            judgeval_logger.warning("Customer ID is already set, skipping.")
+            return
+
         if span and span.is_recording():
             set_span_attribute(span, AttributeKeys.JUDGMENT_CUSTOMER_ID, customer_id)
+            self.customer_id.set(customer_id)
```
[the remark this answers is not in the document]
alanzhang25: I think this is wrong? We should only set customer_id if there is current span?

src/judgeval/tracer/__init__.py
```diff
@@ -247,14 +250,23 @@ def get_tracer(self):
...
+
     def get_span_processor(self) -> JudgmentSpanProcessor:
         """Get the internal span processor of this tracer instance."""
         return self.judgment_processor
 
     def set_customer_id(self, customer_id: str) -> None:
         span = self.get_current_span()
+
+        if self.get_current_customer_id():
+            judgeval_logger.warning("Customer ID is already set, skipping.")
```
abhishekg999, line 261: i think logger debug here if the user decides to disable monitoring we shouldn't dump logs when it shouldnt do anything

src/judgeval/tracer/__init__.py
```diff
@@ -247,14 +250,23 @@ def get_tracer(self):
     def get_current_agent_context(self):
         return self.agent_context
```
abhishekg999, line 238: (question) why does agent context not do a .get() but customer_id does? my understanding is that both of these are context vars

src/judgeval/tracer/__init__.py
```diff
@@ -318,6 +330,17 @@ def record_instance_state(self, record_point: Literal["before", "after"], span):
...
             )
 
+    @dont_throw
+    def add_customer_id_to_span(self, span):
+        customer_id = self.get_current_customer_id()
+        if customer_id:
+            set_span_attribute(span, AttributeKeys.JUDGMENT_CUSTOMER_ID, customer_id)
+
+    @dont_throw
+    def inject_judgment_context(self, span):
```
abhishekg999: nice!

src/judgeval/tracer/__init__.py
alanzhang25: This is the context itself so it seems like the uses of this then do .get or .set?

i think this might be pointless lol

src/judgeval/tracer/__init__.py
alanzhang25: sure

src/judgeval/tracer/__init__.py
```diff
@@ -234,17 +237,46 @@ def get_tracer(self):
...
+            )
+            return
+
+        if self.get_current_customer_context().get():
+            judgeval_logger.warning("Customer ID is already set, skipping.")
+            return
+
         if span and span.is_recording():
             set_span_attribute(span, AttributeKeys.JUDGMENT_CUSTOMER_ID, customer_id)
+            self.get_current_customer_context().set(customer_id)
```
abhishekg999, line 266: this should return a token that lets us reset to the proper state in async environments, is this fine to skip?

PR #623, opened 2025-11-07
stage: v1 wip

abhishekg999:


src/judgeval/v1/scorers/prompt_scorer/prompt_scorer_factory.py
```diff
@@ -0,0 +1,110 @@
...
+    def __init__(
+        self,
+        client: JudgmentSyncClient,
+        is_trace: bool,
+    ):
+        self._client = client
+        self._is_trace = is_trace
+
+    def get(self, name: str) -> PromptScorer:
+        cache_key = (name, self._client.organization_id, self._client.api_key)
```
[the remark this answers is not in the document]
abhishekg999: its unique by name but adding anyways

scripts/api_generator_v1.py
```diff
@@ -0,0 +1,468 @@
...
+
+
+if __name__ == "__main__":
+    import os
+
+    os.makedirs("src/judgeval/v1/internal/api", exist_ok=True)
+
+    generate_api_types()
+    api_code = generate_api_file()
+    print(api_code)
```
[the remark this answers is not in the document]
abhishekg999: @propel-code-bot, dont force DRY principles. DRY often forces code composition that doesnt make sense when things may "seem" identical but in turn are very different, making it hard for future refactors. In this case, the v1 api is truly different, and even though at the moment the client is identical, it reserves the right to implement this in a different pattern.

src/judgeval/v1/datasets/dataset_factory.py
```diff
@@ -0,0 +1,29 @@
...
+from judgeval.v1.data.example import Example
+
+
+class DatasetFactory:
+    __slots__ = "_client"
+
+    def __init__(self, client: JudgmentSyncClient):
+        self._client = client
+
+    def get(self, name: str, project_name: str) -> Dataset:
```
adivate2021, line 17: the way we have it in Java, these methods have the actual logic of getting/creating and then the actual class doesn't have any class methods that do this. I think it probably makes sense to normalize it between the different SDKs.

src/judgeval/v1/scorers/prompt_scorer/prompt_scorer.py
```diff
@@ -0,0 +1,96 @@
+from __future__ import annotations
+
+from typing import Any, Dict, Optional
+
+from judgeval.v1.internal.api.api_types import ScorerConfig
+from judgeval.v1.scorers.base_scorer import BaseScorer
+
+
+class PromptScorer(BaseScorer):
```
adivate2021: can we have PromptScorer inherit from APIScorer since it's not a local scorer. I think it'll remove some of the repeated logic in this file also

src/judgeval/v1/scorers/built_in/built_in_factory.py
```diff
@@ -0,0 +1,26 @@
...
+from judgeval.v1.scorers.built_in.faithfulness import FaithfulnessScorer
+from judgeval.v1.scorers.built_in.instruction_adherence import (
+    InstructionAdherenceScorer,
+)
+
+
+class BuiltInScorersFactory:
+    __slots__ = ()
+
+    def answer_correctness(self, threshold: float = 0.5) -> AnswerCorrectnessScorer:
```
adivate2021, line 14: we can also allow you to set the model/threshold/ other fields the same way that the java version returns the builder. Can make this change later when all the SDKs are together

src/judgeval/v1/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+        self._api_url = api_url
+
+        self._internal_client = JudgmentSyncClient(
+            self._api_url,
+            self._api_key,
+            self._organization_id,
+        )
+
+    @cached_property
+    def Tracer(self):
```
adivate2021: think it might be better to use lowercase for all of these. This might be consistent with OpenAI, can double check with what they do

justinsheu (changes requested):
1. add prompts and move resolve_project_id
2. move async_evaluate and trace_async_evaluate sampling rate to Tracer level

PR #639, opened 2025-11-19
Feat: Filter tracer functionality + Judgment Processor

abhishekg999:
Adds filter by Tracer with the new judgment tracer provider, standardized in go and js.

To test try something like this:

```py
from judgeval import Judgeval
from openai import OpenAI

judgeval = Judgeval()

def filter_tracer(*_, **__) -> bool:
    return False

tracer = judgeval.tracer.create(
    project_name="python-test",
    filter_tracer=filter_tracer,
)

initialize() # adds autoinstrumentations
client = tracer.wrap(OpenAI())

@tracer.observe()
def hello():
    r = client.chat.completions.create(
        model="gpt-4o-mini",
        messages=[
            {"role": "user", "content": "Say hello."},
        ],
    )
    return r.choices[0].message.content

if __name__ == "__main__":
    print(hello())
    tracer.force_flush()
```

Try playing with the filter_tracer function. you can also log the various instrumentors. By returning false it disables all other instrumentors. You can try returning true here and see all instrumentors be provided.

### DIFF
--- a/src/judgeval/trainer/fireworks_trainer.py
+++ b/src/judgeval/trainer/fireworks_trainer.py
@@ -1,7 +1,7 @@
 import asyncio
 import json
 from typing import Optional, Callable, Any, List, Union, Dict
-from fireworks import Dataset  # type: ignore[import-not-found, import-untyped]
+from fireworks import Dataset  # type: ignore[import-not-found,import-untyped]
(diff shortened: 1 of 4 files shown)


alanzhang25 (approved):
this is sick!! lgtm

PR #640, opened 2025-11-19
JUD-2948: Dataset Upload Batch

alanzhang25:
## 📝 Summary

- [ ] 1. ...

## ✅ Checklist

### DIFF
--- a/src/judgeval/dataset/__init__.py
+++ b/src/judgeval/dataset/__init__.py
@@ -153,21 +153,17 @@ class Dataset:
         )
         judgeval_logger.info(f"Created dataset {name}")
 
-        dataset_instance = cls(
-            name=name,
-            project_name=project_name,
-            examples=[],
-        )
-        dataset_instance.add_examples(examples, batch_size=batch_size)
-
         if not isinstance(examples, list):
             examples = list(examples)
 
-        return cls(
+        dataset = cls(
             name=name,
             project_name=project_name,
             examples=examples,
         )
+        dataset.add_examples(examples, batch_size=batch_size)
+
+        return dataset
 
     @classmethod
     def list(cls, project_name: str):
--- a/src/judgeval/v1/datasets/dataset_factory.py
+++ b/src/judgeval/v1/datasets/dataset_factory.py
@@ -77,19 +77,14 @@ class DatasetFactory:
         )
         judgeval_logger.info(f"Created dataset {name}")
 
-        dataset = Dataset(
-            name=name, project_name=project_name, examples=[], client=self._client
-        )
-        dataset.add_examples(examples, batch_size=batch_size)
         if not isinstance(examples, list):
             examples = list(examples)
 
-        return Dataset(
-            name=name,
-            project_name=project_name,
-            examples=examples,
-            client=self._client,
+        dataset = Dataset(
+            name=name, project_name=project_name, examples=examples, client=self._client
         )
+        dataset.add_examples(examples, batch_size=batch_size)
+        return dataset
 
     def list(self, project_name: str) -> List[DatasetInfo]:
         datasets = self._client.datasets_pull_all_for_judgeval(


src/judgeval/v1/datasets/dataset.py
```diff
@@ -59,10 +67,35 @@ def add_from_yaml(self, file_path: str) -> None:
...
+
+        if len(examples) > batch_size:
+            judgeval_logger.info(
+                f"Adding {len(examples)} examples in batches of {batch_size}..."
+            )
+            batches = _batch_examples(examples, batch_size)
 
-    def add_examples(self, examples: List[Example]) -> None:
-        if self.client:
+            for i, batch in enumerate(batches, start=1):
```
abhishekg999: this looks good 1 question do we want some sort of console progress indicator here. we already have rich installed maybe using that?

abhishekg999 (approved):

src/judgeval/dataset/__init__.py
```diff
@@ -15,6 +15,12 @@
 from judgeval.data.judgment_types import DatasetKind
 
 
+def _batch_examples(
```
abhishekg999, line 26: can this be a generator this looks expensive for big batches

src/judgeval/env.py
```diff
@@ -48,7 +48,7 @@ def optional_env_var(var_name: str, default: str | None = None) -> str | None:
 
 
 JUDGMENT_NO_COLOR = optional_env_var("JUDGMENT_NO_COLOR")
-JUDGMENT_LOG_LEVEL = optional_env_var("JUDGMENT_LOG_LEVEL", "WARNING")
```
abhishekg999, line 51: info default?

src/judgeval/v1/datasets/dataset.py
```diff
@@ -9,6 +9,14 @@
 
 from judgeval.v1.data.example import Example
 from judgeval.v1.internal.api import JudgmentSyncClient
+from judgeval.logger import judgeval_logger
+
+
+def _batch_examples(
```
abhishekg999, line 24: same here can this be a generator

src/judgeval/env.py
alanzhang25: Yeah I feel we should show more info to users and they can change if want.

src/judgeval/dataset/__init__.py
```diff
@@ -103,24 +116,43 @@ def create(
         project_name: str,
         examples: List[Example] = [],
```
abhishekg999: One more nit, can examples be an Iterable[Example].
Therefore users can pass lists but also generators here.

the usecase is: I think users who use huggingface will be used to using generators since that is how they do their datasets too.

this will require the batch method to be a bit smarter, but will still work. The batch method instead of using [:] will need to iterate N at a time and yield it should be a simple change

src/judgeval/dataset/__init__.py
```diff
@@ -175,21 +207,58 @@ def add_from_yaml(self, file_path: str) -> None:
...
+                BarColumn(),
+                TaskProgressColumn(),
+                TextColumn("[dim]{task.fields[info]}"),
+            ) as progress:
+                task = progress.add_task(
+                    f"Uploading to {self.name}", total=total_batches, info=""
+                )
+
+                for i, batch in enumerate(
+                    _batch_examples(examples, batch_size), start=1
```
abhishekg999: Something is wrong with this logic here:

```py
import dotenv
dotenv.load_dotenv("../.env")


from judgeval.v1 import Judgeval
from judgeval.v1.data.example import Example

judgeval = Judgeval()


dataset = judgeval.datasets.create(name="test", project_name="test", overwrite=True)


def generate_examples(n: int):
    for i in range(n):
        yield Example.create(input=f"input {i}", actual_output=f"output {i}")


examples = list(generate_examples(1000))
dataset.add_examples(examples)
```

https://staging.app.judgmentlabs.ai/org/27c7d380-6cb0-495f-b27c-8c1ff82b72af/project/bb760692-8950-47fa-9988-a87e91313869/datasets/85882286-a65c-45e3-be5b-37d51823e8fd?time_range=all

Only the first 100 records are being uploaded repeatedly.

see the data there is no input > 100

abhishekg999 (changes requested):
some issues with the batching, the console looks good overall

src/judgeval/env.py
abhishekg999: ususally since monitoring is tied in here I think general practice is to keep default log level as warning. 
this matches python logger behavior too.

Also info is meant to be stuff that users can opt into to see more about whats going on, like the number of spans emitted

https://docs.python.org/3/howto/logging.html
@alanzhang25

abhishekg999: log is also not meant to show crutial display stuff its just logs always for error handling and management for information that is not explicitly important, if in datasets you want to explicitly show some text I would say use rich?

src/judgeval/dataset/__init__.py
```diff
@@ -175,20 +213,53 @@ def add_from_yaml(self, file_path: str) -> None:
...
+                    task,
+                    advance=1,
+                    info=f"Batch {batch_num} ({batch_size_actual} examples, {total_uploaded} total)",
+                )
+
+                client.datasets_insert_examples_for_judgeval(
+                    {
+                        "dataset_name": self.name,
+                        "project_name": self.project_name,
+                        "examples": batch,  # type: ignore
```
abhishekg999, line 257: why type ignore this should be properly typed

PR #643, opened 2025-11-29
Isolated Tracer fix

alanzhang25:
## 📝 Summary

- [ ] 1. ...

## ✅ Checklist

### DIFF
--- a/src/judgeval/v1/integrations/claude_agent_sdk/wrapper.py
+++ b/src/judgeval/v1/integrations/claude_agent_sdk/wrapper.py
@@ -15,7 +15,6 @@ from typing import (
     Tuple,
 )
 
-from opentelemetry import trace
 from opentelemetry.trace import set_span_in_context
 
 from judgeval.tracer.keys import AttributeKeys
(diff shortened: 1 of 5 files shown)


src/judgeval/v1/tracer/tracer.py
```diff
@@ -64,3 +66,9 @@ def force_flush(self, timeout_millis: int = 30000) -> bool:
 
     def shutdown(self, timeout_millis: int = 30000) -> None:
         self._tracer_provider.shutdown()
+
+    def get_context(self) -> Context:
```
abhishekg999: let this live on base tracer if possible and can remove the type ignore

abhishekg999: basically base tracer should store a tracer and a context api reference so any tracer can use this if possible,

if not and this seems unclean then let the typehint of wrap expect a tracer not a base tracer

src/judgeval/v1/integrations/claude_agent_sdk/wrapper.py
```diff
@@ -15,7 +15,7 @@
     Tuple,
 )
 
-from opentelemetry import trace, context as otel_context
+from opentelemetry import trace
```
abhishekg999: i see the trace import still exists is this used? ideally trace api shoudlnt be used but its possible it will work if explicitly using a context /. tracer from judgeval tracer

PR #747, opened 2026-06-10
Aaryan/offline tests v1

adivate2021:


adivate2021:
Synced the SDK to the current judgment-mono server contracts (verified each against `aaryan/offline-tests-judgeval-server`):

**Paginated items reads** (`a5a99ff6`) — `GET /test-runs/{id}/items` is now keyset-paginated by `example_id` (limit default/max 200, `has_more` + `next_cursor`), with per-scorer `str_value`/`reason`/`error` truncated to 128 chars server-side. `fetch_items` now loops with the cursor until `has_more` is false or `next_cursor` is null, passes `limit=200` explicitly, tolerates old servers that omit the pagination fields, and bounds the loop like `fetch_examples`. It uses the raw `_request` helper (same pattern as the success PATCH) until a client regen exposes the query params. Docstrings note the 128-char truncation for `pass_condition_fn` authors.

**`versioned_results` removed** (`23dc3a9e`) — the server dropped the field from `CreateTestRunRequest` along with the multi-version guard (queue refs carry pinned judge versions now). Removed the payload key, the `has_duplicate_judges` auto-set, and the public parameter; pinning the same judge twice just passes through.

**Ad-hoc evaluation restored** (`8c221e8f`) — the server reverted `/eval-results` + `/eval-results/examples` to pure legacy semantics and `/experiments/{runId}` falls back to the legacy experiment fetch, so both earlier SDK mitigations were reverted: the hosted-runner 501 fast-fail is gone (eval-queue submit + experiments polling restored, hosted-scorers docstrings back), and the local-results 422 redirect to `client.offline_tests` is removed. The regenerated-client adaptations (`OfflineTestsLocalScorerResult`, `judge_name`-keyed scorer rows, success passthrough) and the orthogonal tracer-restoration fix are kept.

**Error contract** — verified no change needed: server errors are still `{error, message}` and the client's `detail → message → error` fallback handles them; `map_judgment_api_error` paths untouched.

Checks: 610 tests passing, ruff and mypy clean.

src/judgeval/offline_tests/offline_test_runner.py
```diff
@@ -0,0 +1,853 @@
...
+
+    # ------------------------------------------------------------------ #
+    #  Orchestration                                                     #
+    # ------------------------------------------------------------------ #
+
+    def run(
+        self,
+        test_config: TestConfig,
+        agent_function: Optional[AgentFunction] = None,
+        judge_versions: Optional[List[Dict[str, Any]]] = None,
```
justinsheu: can we make this typign stricter

justinsheu (changes requested):
1. db schema json doesn't seem to have any validation client or server side? should probably add to both ideally. also on sdk is there a way we can have a better typed way to define the schema (e.g. stricter type than just dict[str, Any] or maybe a dataclass)?
2. can we make it clear what happens when you don't pass in agent_function? it seems like it just runs the eval over the existing examples in the dataset rn?
3. do we have equivalent for pass_condition_fn on platform? or like on platform is there a way to indicate success/failure?

src/judgeval/datasets/dataset_factory.py
```diff
@@ -15,6 +15,65 @@
...
+
+class DatasetSchemaProperty(TypedDict):
+    """A single dataset column declaration.
+
+    `type` is a JSON Schema primitive (`"string"`, `"integer"`,
+    `"number"`, `"boolean"`, `"array"`, `"object"`) or the Judgment
+    pointer type `"trace"` (value is a trace id).
+    """
+
+    type: str
```
justinsheu: this can be union of literals right? the options for type is bounded no?

src/judgeval/datasets/dataset_factory.py
```diff
@@ -278,24 +343,28 @@ def create(
...
                     "client.datasets.create(), or provide `examples` to infer one."
                 )
-            schema = infer_schema_from_examples(examples)
+            resolved_schema = infer_schema_from_examples(examples)
             judgeval_logger.info(
                 f"No schema provided for dataset {name}; inferred one from "
                 f"{len(examples)} example(s)"
             )
+        else:
+            resolved_schema = dict(schema)
```
justinsheu, line 375: lol why are we converting back to generic dict before passing into validate_dataset_schema

justinsheu (changes requested):

justinsheu (approved):
lgtm

PR #749, opened 2026-06-23
Fix: Install as global tracer provider controls otel context 

abhishekg999:
* fix: install as global tracer provider controls otel context as well

* format


justinsheu (approved):
lgtm

PR #756, opened 2026-07-17
JUD-8016: Add public JQL to the Python SDK

JJwilkin:
## Summary

- add the dependency-free `judgeval.jql` public builder surface
- add synchronous `Judgeval.query()`, `Judgeval.present()`, and `Judgeval.discover()` methods
- generate operation and discovery metadata from the DAL OpenAPI `x-jql` contract
- generate public response types from the judgeval-server public OpenAPI contract
- preserve typed error codes, hints, and `Retry-After`
- keep tenant identifiers in authenticated client context, never the query body
- retain Python 3.10 support without adding a private DAL or JQL runtime dependency
- regenerate JQL artifacts automatically from public-safe snapshots during Hatch builds
- record deterministic hashes of the complete public IR and transport contracts
- fail PR and release CI if generated output is stale or differs from `judgment-mono/main`

## Impact

Python users can construct and execute the same canonical public JQL IR as the TypeScript SDK through their existing Judgeval client configuration. Package builds remain credential-free, while a separate CI job uses read-only cross-repository access to prove parity with the private monorepo source.

## Testing

- `uv run --frozen pytest src/tests/jql -q` — 6 passed
- full unit suite — 623 passed
- `uv run --frozen ruff format --check .` and `uv run --frozen ruff check .` — passed
- `uv run --frozen mypy src/judgeval` — passed across 196 source files
- `uv build` — sdist and wheel both regenerated JQL artifacts successfully
- GitHub Actions — all 18 checks passed, including generated-contract freshness and live `judgment-mono/main` parity

## Dependency and rollout

- depends on the merged JUD-8016 monorepo stack, ending with `JudgmentLabs/judgment-mono#3068`
- runtime availability remains guarded by the per-organization `public_jql` feature flag
- staging, production, and manual release workflows recheck monorepo parity before publishing

## Checklist

- [x] Docs updated in `README.md`
- [x] Release notes will be generated by the normal staging-to-main SDK release flow


.github/workflows/release.yaml
```diff
@@ -17,11 +17,29 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Checkout canonical JQL sources
```
SamGearou, line 24: **Blocking: this publishes the entire private monorepo to PyPI.**

`.jql-source/judgment-mono` is checked out *inside the repo working tree*, and `python -m build` runs later in the same job (line 119). Hatchling's sdist target includes every file in the project root that isn't VCS-ignored — `.jql-source/` is neither in `.gitignore` nor in the `exclude` list in `pyproject.toml`, so it ends up in the tarball.

Reproduced on this branch:

```
$ mkdir -p .jql-source/judgment-mono/services
$ echo "PRIVATE MONOREPO CONTENT" > .jql-source/judgment-mono/services/secret.ts
$ uv build --sdist
$ tar -tzf dist/judgeval-0.0.0.tar.gz | grep jql-source
judgeval-0.0.0/.jql-source/judgment-mono/services/secret.ts
```

The `git diff --exit-code` guards don't catch it because the directory is untracked, not modified.

Same problem in `release-staging.yaml:21` and `manual-release.yaml:44`.

Suggested fix — do all three:
1. `path: ${{ runner.temp }}/judgment-mono` so the checkout is never inside the package root;
2. add `.jql-source/` to `.gitignore`;
3. add an explicit `[tool.hatch.build.targets.sdist]` with a positive `include` list rather than relying on the implicit "everything not ignored" default.

.github/workflows/ci.yaml
```diff
@@ -10,6 +10,77 @@ jobs:
...
+          git diff --exit-code -- \
+            src/judgeval/jql/_generated_contract.py \
+            src/judgeval/jql/_generated_transport.py
+
+      - name: Build distributions
+        run: |
+          pip install uv
+          uv build
+
+      - name: Check package build preserves generated output
```
SamGearou: This step can never fail. The step at line 26 already ran `python scripts/generate_jql.py` (from the checked-in snapshots) and diffed; `uv build` invokes the hatch hook which runs *the exact same command against the exact same inputs*. The second diff is a tautology.

What would actually be worth asserting here is that the built artifacts contain the generated modules and nothing unexpected — e.g. `tar -tzf dist/*.tar.gz` / `unzip -l dist/*.whl` and check the file list. That would also have caught the `.jql-source` leak flagged on `release.yaml`.

src/judgeval/judgeval.py
```diff
@@ -179,6 +186,63 @@ def offline_tracer(
...
+        kind: "DiscoveryKind",
+        *,
+        limit: Optional[int] = None,
+        **options: Any,
+    ) -> "JqlQueryResponse":
+        """Discover project-scoped judges, fields, models, and related values."""
+        from judgeval.jql import discovery
+
+        if limit is not None:
+            options["limit"] = limit
```
SamGearou: `limit` is sent twice, in two different places:

```python
if limit is not None:
    options["limit"] = limit
return self.query(discovery(kind, **options), limit=limit)
```

produces `{"query": {"op": "discovery", "kind": "judges", "limit": 25}, "limit": 25}` — once inside the IR node (`DiscoveryQuery.limit` is a real field in `jql-ir.openapi.json`) and once in the request envelope. Two independent limit semantics on one request is at best redundant and at worst silently conflicting, and nothing in the SDK reconciles them.

Pick one. Given `discover()` is sugar over `query()`, dropping the `options["limit"] = limit` line and letting the envelope carry it is the smaller change — but if the discovery IR limit is the one the server honours, do the inverse and don't forward `limit=` to `query()`.

src/judgeval/judgeval.py
```diff
@@ -179,6 +186,63 @@ def offline_tracer(
...
+        return cast(
+            "JqlQueryResponse",
+            self._internal_client._request(
+                "POST",
+                url_for(f"/v1/projects/{project_id}/query", self._api_url),
+                payload,
+            ),
+        )
+
+    def present(
```
SamGearou, line 202: Asymmetric with `query()` directly above: `query()` takes `QueryInput` and normalises through `to_json()`, but `present()` takes a bare `Dict[str, Any]` and passes it through untouched.

This works only because `chart()`/`table()` happen to return plain dicts, but it means:
- passing a `QueryBuilder` (an easy mistake, since that's what `query()` accepts) type-checks as `Dict[str, Any]`? no — it fails mypy, but at runtime any dict-ish input goes straight onto the wire unvalidated and un-deep-copied;
- the caller's dict is aliased into the payload, unlike everywhere else in this module, which is scrupulous about `deepcopy`.

Make it `query: "QueryInput"` and run it through `to_json()` like `query()` does.

src/judgeval/judgeval.py
```diff
@@ -179,6 +186,63 @@ def offline_tracer(
...
+        """Discover project-scoped judges, fields, models, and related values."""
+        from judgeval.jql import discovery
+
+        if limit is not None:
+            options["limit"] = limit
+        return self.query(discovery(kind, **options), limit=limit)
+
+    def _require_jql_project_id(self) -> str:
+        if not self._project_id:
+            raise ValueError(
```
SamGearou: Nit, but user-facing: `f"Project '{self._project_name}' must resolve before running JQL."` describes an internal invariant rather than what the user did wrong. `_project_id` is resolved eagerly in `__init__`, and when resolution fails the constructor only logs a warning (line 120) — so the actual condition here is "the project doesn't exist / isn't visible to these credentials".

Something like `f"Project '{self._project_name}' was not found for this organization; JQL queries require a resolved project."` is more actionable. Also consider raising a `judgeval.exceptions` type rather than bare `ValueError`, since the rest of this PR is about giving callers typed, catchable errors.

src/judgeval/jql/__init__.py
```diff
@@ -0,0 +1,470 @@
...
+    def last(self, window: str) -> "QueryBuilder":
+        return self._replace(time={"last": window})
+
+    def since(self, since: str) -> "QueryBuilder":
+        return self._replace(time={"since": since})
+
+    def between(self, start: str, end: str) -> "QueryBuilder":
+        return self._replace(time={"between": [start, end]})
+
+    def pipe(self) -> "PipelineBuilder":
```
SamGearou, line 349: `pipe()` silently discards a select that was already applied:

```python
>>> traces().ids().pipe().to_json()
{'op': 'query', 'source': 'traces', 'pipe': []}
```

The `ids()` is gone with no warning. Since `SourceQuery` is `additionalProperties: false` and `select`/`pipe` are mutually exclusive in practice, silently dropping is the wrong default — raise `ValueError("pipe() cannot follow a select; call pipe() before rows()/ids()/count()/...")` instead of `spec.pop("select", None)`.

Separately: `to_json()` on a `PipelineBuilder` unconditionally emits `"pipe": []` even with zero stages, so `spans().pipe().to_json()` sends an empty pipe rather than a plain source query. Worth omitting the key when `_stages` is empty.

src/judgeval/jql/__init__.py
```diff
@@ -0,0 +1,470 @@
...
+    def count(self, by: Optional[str] = None) -> "QueryBuilder":
+        return self._select(_compact({"op": "count", "by": by}))
+
+    def recent(self, n: int) -> "QueryBuilder":
+        return self._select({"op": "recent", "n": n})
+
+    def top(self, n: int, by: str) -> "QueryBuilder":
+        return self._select({"op": "top", "n": n, "by": by})
+
+    def ranked(self, **options: Any) -> "QueryBuilder":
```
SamGearou: The `**options` splat comes *after* the literal keys here, so caller kwargs silently overwrite the op discriminator:

```python
>>> traces().ranked(op='evil').to_json()
{'op': 'query', 'source': 'traces', 'select': {'op': 'evil'}}
```

Same pattern in `PipelineBuilder.pick()` (line 348) and `discovery()` (line 397, where `**options` can also clobber `kind`).

Note this is the opposite ordering from `chart()`/`table()` (lines 317/320), which splat options *first* and are therefore safe. Make it consistent — put the splat first everywhere so `op`/`kind`/`query` always win.

src/judgeval/jql/__init__.py
```diff
@@ -0,0 +1,470 @@
...
+        )
+
+    def trend(
+        self, *, metric: Optional[str] = None, bucket: Optional[str] = None
+    ) -> "QueryBuilder":
+        return self._select(
+            _compact({"op": "trend", "metric": metric, "bucket": bucket})
+        )
+
+    def chart(self, **options: Any) -> JsonObject:
```
SamGearou: `chart(**options)` / `table(**options)` erase all type information for fields the IR marks required. Per `jql-ir.openapi.json`, `ChartQuery.required = [op, chart_type, title, x_axis, y_axis, query]` and `TableQuery.required = [op, title, columns, query]`, and both are `additionalProperties: false`.

So `spans().chart()` builds happily and fails server-side with a 422, and a typo like `chart_typ=` is indistinguishable from a valid call at author time. For a surface whose entire justification is "generated from the canonical contract", the two most structured nodes are the least typed things in the module.

At minimum give these explicit keyword parameters (`chart_type`, `title`, `x_axis`, `y_axis`, `series_by=None`, `description=None`) — the codegen already has the schema to derive them from, and it would let `mypy` catch these instead of the server.

SamGearou (commented):
Detailed review of the public JQL surface. The overall shape (dependency-free builders + generated contract + CI parity) is good, but there is one release-blocking packaging issue and a handful of logic bugs in the builders and the generator. Findings are left as inline comments.

The one that must be fixed before this can ship: checking `judgment-mono` out into `.jql-source/` inside the working tree means `python -m build` packages the entire private monorepo into the published sdist. I reproduced this locally — details in the comment on `release.yaml`.

src/judgeval/jql/__init__.py
```diff
@@ -0,0 +1,470 @@
...
+            "op": op,
+            "agg": dict(agg),
+            "cmp": cmp,
+            "value": value,
+            "where": dict(where) if where is not None else None,
+        }
+    )
+
+
+def over_spans(
```
SamGearou, line 219: `_over` types `cmp` as `Literal["eq", "ne", "gt", "gte", "lt", "lte"]`, but all three public wrappers widen it to `str` and then suppress the resulting error with `# type: ignore[arg-type]` (lines 198, 204, 208).

That throws away the only type safety the literal was providing, at exactly the boundary where it matters — the public API. Hoist the alias and use it on the wrappers:

```python
Cmp = Literal["eq", "ne", "gt", "gte", "lt", "lte"]

def over_spans(agg: Mapping[str, Any], cmp: Cmp, value: float, where: Optional[Filter] = None) -> JsonObject:
    return _over("over_spans", agg, cmp, value, where)
```

and delete all three ignores.

src/judgeval/jql/__init__.py
```diff
@@ -0,0 +1,470 @@
...
+    agg: Mapping[str, Any], cmp: str, value: float, where: Optional[Filter] = None
+) -> JsonObject:
+    return _over("over_traces", agg, cmp, value, where)  # type: ignore[arg-type]
+
+
+def over_scores(agg: Mapping[str, Any], cmp: str, value: float) -> JsonObject:
+    return _over("over_scores", agg, cmp, value)  # type: ignore[arg-type]
+
+
+def at_least(
```
SamGearou, line 235: Inconsistent null-handling for `where` across the module:

- `_over` (line 190) uses `dict(where) if where is not None else None`
- `at_least` (here), `agg_expr` (line 228) and `_query` (line 377) use `if where` / `if filter` — **truthiness**

So an empty filter is preserved in one place and silently dropped in the other:

```python
>>> over_spans({'func':'sum'}, 'gt', 1, where={})
{'op': 'over_spans', ..., 'where': {}}
>>> at_least(1, 'spans', where={})
{'op': 'at_least', 'k': 1, 'of': 'spans'}     # where vanished
```

An empty mapping is a caller bug either way, but the two behaviours shouldn't disagree. Standardise on `is not None` (which surfaces the bad filter to the server) everywhere.

src/judgeval/jql/__init__.py
```diff
@@ -0,0 +1,470 @@
...
+def to_json(query: QueryInput) -> JsonObject:
+    return (
+        query.to_json()
+        if isinstance(query, (QueryBuilder, PipelineBuilder))
+        else deepcopy(query)
+    )
+
+
+# Python-safe names are primary; aliases retain the canonical combinator vocabulary.
+all = all_
```
SamGearou, line 596: Exporting `all` and `any` in `__all__` (lines 427/430) means `from judgeval.jql import *` shadows two builtins in the caller's namespace. That's a nasty, hard-to-debug footgun for a public SDK — `all(x > 0 for x in xs)` starts raising `TypeError` several lines away from the import.

The `all_` / `any_` names are already exported and are the idiomatic Python spelling for this exact collision. I'd keep the module-level aliases for `judgeval.jql.all(...)` attribute access if you want the canonical vocabulary available, but drop `"all"` and `"any"` from `__all__` so star-import can't clobber builtins.

(Also note the alias assignments shadow the builtins inside this module itself for anything defined after line 413.)

scripts/generate_jql.py
```diff
@@ -0,0 +1,183 @@
...
+
+    return {
+        "openapi": "3.1.0",
+        "info": {"title": "Public JQL IR", "version": "1"},
+        "paths": {},
+        "components": {"schemas": included},
+    }
+
+
+def python_type(schema: dict[str, Any]) -> str:
```
SamGearou, line 75: `python_type` has three gaps that are latent today only because the two response schemas it currently sees are trivially simple:

```python
>>> python_type({'$ref': '#/components/schemas/TimeSpec'})
'Any'                                    # refs silently degrade
>>> python_type({'anyOf': [{'type':'string'},{'type':'integer'},{'type':'null'}]})
'Union[str, int, Any]'                   # 'null' -> Any, collapsing the union
>>> python_type({'type': ['string', 'null']})
TypeError: cannot use 'list' as a dict key (unhashable type: 'list')
```

The third is the concerning one: `"type": [...]` is standard OpenAPI 3.1 (which is what the contract declares — `"openapi": "3.1.0"`), so the generator will hard-crash the moment judgeval-server emits a nullable field in that form. That's a broken release build, not a degraded type.

Please handle `$ref` (emit the referenced class name / `Any` with a comment), map `"null"` to `None` inside unions, and normalise `type` to a list before the lookup.

scripts/generate_jql.py
```diff
@@ -0,0 +1,183 @@
...
+
+    public_schemas = public_document["components"]["schemas"]
+    classes = []
+    for name in ("PublicJqlQueryResponse", "PublicJqlPresentationResponse"):
+        schema = public_schemas[name]
+        fields = "\n".join(
+            f"    {field}: {python_type(field_schema)}"
+            for field, field_schema in schema["properties"].items()
+        )
+        classes.append(f"class {name.removeprefix('Public')}(TypedDict):\n{fields}")
```
SamGearou: The generator ignores the schema's `required` list entirely and always emits a `total=True` `TypedDict`.

It happens to be correct right now (both `PublicJqlQueryResponse` and `PublicJqlPresentationResponse` mark all four properties required), but the first optional field added upstream will produce a `TypedDict` that lies — `mypy` will accept `response["foo"]` on a key the server may not send, and reject constructing the dict without it.

Since the target is Python ≥3.10, `typing.NotRequired` isn't available from `typing` until 3.11 — use `typing_extensions`, or emit two `TypedDict` bases (`class X(XRequired, total=False)`).

scripts/generate_jql.py
```diff
@@ -0,0 +1,183 @@
...
+    "{jql_contract_hash}"
+)
+SUPPORTED_OPS = (
+{operation_members}
+)
+DiscoveryKind = Literal[
+{literal_members}
+]
+'''
+    OUTPUTS[0].write_text(contract_source)
```
SamGearou: `Path.write_text` here (and at lines 130, 172, 175) uses the platform default encoding and default newline translation. On Windows that writes CRLF, which makes the generated files byte-differ from what CI produced and trips the `git diff --exit-code` parity gates for anyone developing there.

Given the whole point of this script is deterministic, byte-reproducible output, pin it: `write_text(source, encoding="utf-8", newline="\n")`.

hatch_build.py
```diff
@@ -0,0 +1,18 @@
...
+import sys
+from pathlib import Path
+
+from hatchling.builders.hooks.plugin.interface import BuildHookInterface
+
+
+class CustomBuildHook(BuildHookInterface):
+    """Regenerate deterministic JQL artifacts before building a distribution."""
+
+    def initialize(self, version: str, build_data: dict[str, object]) -> None:
```
SamGearou, line 13: This hook runs unconditionally on *every* build, including when an end user installs from the sdist (`pip install --no-binary judgeval`, or any platform without a matching wheel). At that point `python scripts/generate_jql.py` executes on the user's machine and `check=True` turns any hiccup into a hard install failure — for codegen whose output is already checked in and already verified by three separate CI gates.

The sdist does ship `scripts/` and the 4.4k-line contract JSON so it currently works, but that's ~200KB of build-only input shipped to every user for no runtime benefit.

Gate it to the repo/dev case, e.g. skip when the source tree isn't a git checkout, or key it off an env var that only CI and `uv build` set. `version` and `build_data` are both unused — `version` is exactly the hook parameter meant for this kind of branching.

src/tests/jql/test_jql.py
```diff
@@ -0,0 +1,209 @@
...
+from judgeval.internal.api.api_client import JudgmentSyncClient, _handle_response
+from judgeval.jql import (
+    agg_expr,
+    col,
+    eq,
+    gte,
+    spans,
+    traces,
+)
+from judgeval.jql._generated_contract import SUPPORTED_OPS
```
SamGearou, line 29: This test doesn't test what its name says. It asserts a hand-copied 55-element literal equals `SUPPORTED_OPS` — i.e. it restates the generated file's contents against itself. It never touches "the builder surface", so it cannot catch the case it's presumably guarding: a builder function emitting an op the server doesn't support.

It also makes every legitimate upstream contract change a two-file edit for zero signal, in a repo that already has three CI jobs enforcing the generated file matches its source.

What would actually be worth asserting:

```python
def test_every_builder_op_is_supported() -> None:
    emitted = {
        traces().to_json()["op"],
        eq("a", 1)["op"],
        ...
    }
    assert emitted <= set(SUPPORTED_OPS)
```

Better still, drive it off the `__all__` list so new builders are covered automatically.

src/judgeval/jql/_generated_contract.py
```diff
@@ -0,0 +1,75 @@
+"""Generated from DAL OpenAPI x-jql metadata; do not edit."""
+
+from typing import Literal
+
+JQL_IR_SCHEMA_SHA256 = (
```
SamGearou: `JQL_IR_SCHEMA_SHA256` and `PUBLIC_JQL_OPENAPI_SHA256` (in `_generated_transport.py`) are computed, written, and then never read — nothing in `src/judgeval` imports either constant. Same for `SUPPORTED_OPS`, whose only consumer is the tautological test in `test_jql.py`.

So the PR description's "record deterministic hashes of the complete public IR and transport contracts" is true of the file but has no effect: the builders perform zero validation against the contract, and a drifted client fails at the server rather than locally.

Either wire them up — e.g. have the builders assert their `op` is in `SUPPORTED_OPS`, and/or send the IR hash as a header so the server can reject a stale client — or drop the hashes, since a constant nobody reads is just another thing to keep in sync.

src/judgeval/internal/api/api_client.py
```diff
@@ -31,9 +31,25 @@ def _handle_response(r: Response) -> Any:
             detail = (
                 body.get("detail") or body.get("message") or body.get("error") or ""
             )
+            code = body.get("error")
```
SamGearou, line 35: Minor, but worth a comment in the generator: `body.get("error")` feeds both `detail` (as the third fallback, line 31) and `code` here. For an error body of `{"error": "JQL_RATE_LIMITED"}` with no `detail`/`message`, the user gets `JudgmentAPIError("429: JQL_RATE_LIMITED")` with `.code == "JQL_RATE_LIMITED"` — the machine-readable code doubling as the human-readable message.

That's a reasonable fallback, but it's non-obvious enough that the next person reading `scripts/api_generator_v1.py` will wonder if it's a copy-paste bug. One line of comment in the generated source would settle it.

src/judgeval/judgeval.py
JJwilkin: Fixed in 885f399d. `discover()` now sends `limit` only in the request envelope; the discovery IR no longer receives a duplicate value. Added an exact payload regression test.

src/judgeval/jql/__init__.py
JJwilkin: Fixed in 885f399d. `pipe()` now raises if a select is already present instead of silently discarding it, and an empty pipeline omits the `pipe` key. Both cases have regression tests.

src/judgeval/jql/__init__.py
JJwilkin: Fixed in 885f399d. `chart()` and `table()` on both builder types now expose explicit required keyword parameters and typed optional fields. `chart_type` is constrained to the contract literal values and presentation fields are typed mappings.

src/judgeval/jql/__init__.py
JJwilkin: Fixed in 885f399d. Builder construction now goes through `_node()`, which writes the canonical discriminator after caller-provided values. This covers `ranked`, `pick`, `discovery`, and the other kwargs-based builders, with override regression tests.

src/judgeval/jql/__init__.py
JJwilkin: Fixed in 885f399d. Hoisted the comparison literal to the public `Cmp` alias, applied it to all three wrappers, and removed the type ignores. Mypy passes across all 196 source files.

.github/workflows/release.yaml
JJwilkin: yikes good find. addressed with the following:
- the SDK and judgment-mono are now checked out as sibling directories in parity/release jobs, 
- `.jql-source/` is ignored and excluded, and the sdist uses a positive include list. 
- CI also plants a private-source directory and validates the wheel/sdist file manifests removes it before upload. 
-  reproduced this locally and confirmed the sentinel is absent from both artifacts.

src/judgeval/judgeval.py
JJwilkin: changed `present()` to use `QueryInput` and runs it through `to_json()`

src/judgeval/jql/__init__.py
JJwilkin: Fixed in 885f399d. `all` and `any` remain available for qualified attribute access, but were removed from `__all__`; `all_` and `any_` remain the star-import-safe exports. Added a regression test.

src/judgeval/judgeval.py
JJwilkin: Fixed in 885f399d. The message now explains that the project was not found for the configured organization, and the path raises a typed `JudgmentProjectNotFoundError` (a `ValueError` subclass for compatibility).

src/judgeval/jql/__init__.py
JJwilkin: Fixed in 885f399d. `_query`, `at_least`, and `agg_expr` now use `is not None`, matching `_over`, so empty mappings are preserved consistently. Added regression coverage.

scripts/generate_jql.py
JJwilkin: Fixed in 885f399d. Generated response types now honor `required`; mixed schemas emit a required TypedDict base plus a `total=False` subclass for optional keys, avoiding a Python 3.11-only dependency. Covered by a compile test.

scripts/generate_jql.py
JJwilkin: Fixed in 885f399d. The generator now handles OpenAPI 3.1 type arrays, maps null to `None` in unions, and handles `$ref` as `Any` with an emitted reference comment so generated code stays importable. Added focused generator and compile tests.

.github/workflows/ci.yaml
JJwilkin: replaced it with `scripts/check_distribution_contents.py`, which checks both artifacts against an allowlist, requires the generated modules, and catches a for leaked artifacts like above ^

scripts/generate_jql.py
JJwilkin: Fixed in 885f399d. All contract reads/writes now pin UTF-8, and generated/snapshot writes pin `newline="\n"` for byte-stable output across platforms.

src/judgeval/jql/_generated_contract.py
JJwilkin: Fixed in 885f399d. Removed both unused hash constants. `SUPPORTED_OPS` is now a production dependency used by the central `_node()` builder path, with tests proving unsupported operations fail locally.

src/tests/jql/test_jql.py
JJwilkin: Fixed across 8aadd2a7 and 885f399d. The hand-copied literal was replaced with builder-surface introspection, and production builders now validate every emitted discriminator through `_node()` against `SUPPORTED_OPS`. Added a failure-path test as well.

src/judgeval/internal/api/api_client.py
JJwilkin: Fixed in 885f399d. Added the explanatory comment to both the API generator source and the checked-in generated client so the intentional human-readable fallback is clear.

hatch_build.py
JJwilkin: Fixed in 885f399d. The hook now regenerates only in a git checkout. The sdist carries checked-in generated modules but not the codegen inputs, and building the wheel back from the sdist skips repo-only codegen successfully.

abhishekg999 (approved):

.github/workflows/release.yaml
JJwilkin: Further hardening in df43d588: dropped hatch's `artifacts` force-include, which was shipping gitignored `.py` files under `src/judgeval/` into both distributions (verified with a probe file: present with the setting, absent without; artifact file lists on a clean tree are byte-identical otherwise). So the manifest guarantee is no longer only "nothing outside the package dir" — build-time strays inside the package dir no longer ship either.

.github/workflows/ci.yaml
JJwilkin: Tightened the allowlist in df43d588 after probing `check_distribution_contents.py` for dodges: the wheel check now requires `*.dist-info` as the root path segment (previously the substring anywhere passed, e.g. `evil/.dist-info/x`), and the sdist check requires paths exactly under `src/judgeval/` instead of the `src/judgeval` prefix (which also matched `src/judgeval_evil.py`).

src/judgeval/jql/__init__.py
JJwilkin: Strengthened in 2b363217: a conflicting caller-supplied `op` now raises `ValueError` instead of being silently overwritten — silently winning felt like the same footgun pointed the other way. A redundant `op` equal to the canonical one is still accepted. And with d255e969 giving `ranked()`/`pick()` explicit signatures, `op=` there fails as a `TypeError` at call time.

src/judgeval/jql/__init__.py
JJwilkin: Extended the same principle in 2b363217: a second select now raises instead of silently replacing the first — `traces().ids().count()` was quietly discarding the `ids()`, which is the same silent-discard class as pipe-after-select. Select-vs-select and select-vs-pipe conflicts now behave consistently, with regression tests.

src/judgeval/jql/__init__.py
JJwilkin: Follow-through completed in d255e969: `ranked()` and `pick()` now have explicit contract-typed keyword parameters as well (`by`/`pick`/`within` and `by`/`n`/`per`/`reverse` per RankedSelect/PickStage), so kwarg typos on every structured builder now fail locally instead of as server-side 422s.

JJwilkin:
Pushed five hardening commits on top of 885f399d, following a second pass over the public surface in the spirit of the review:

- **df43d588** — packaging: drop hatch's `artifacts` force-include (gitignored `.py` under `src/judgeval/` was shipping in both artifacts) and tighten the distribution manifest allowlist (`*.dist-info` root-segment only, exact `src/judgeval/` prefix).
- **d255e969** — `ranked()`/`pick()` get explicit contract-typed keyword parameters, same treatment as `chart()`/`table()`.
- **2b363217** — conflicting `op` kwargs and second selects now raise instead of silently discarding caller intent, consistent with the pipe-after-select fix.
- **f7c775c4** — `descendant_of()`/`ancestor_of()` depth typed as `Optional[Literal[1]]` (`HierarchyDepth`), matching the IR's `const 1 | null` instead of advertising arbitrary ints.
- **934881ee** — JQL calls now route errors through `map_judgment_api_error`, so 422 → `JudgmentValidationError` and 409 → `JudgmentConflictError` like the rest of the SDK; `code`/`hint`/`retry_after_seconds` preserved.

Since @abhishekg999 approved at 885f399d: the behavior changes worth a re-look are the two new `ValueError`s in 2b363217 (previously silent success) and the stricter `ranked()`/`pick()` signatures in d255e969 (unknown kwargs now fail fast). Verification at head: 636 unit tests pass, mypy clean across 196 files, ruff clean, build + manifest checks green, generated modules byte-stable.

